PR #4, opened 2020-10-02
Replace Travis with Github Actions

michel2323:
This largely replaces Travis with Github Actions. It runs

* pkg> test
* the tests that were previously in `.travis.yml`

Github Actions should be free for public repositories and they allow to do self-hosted runners and run tests with the GPU enabled. 

I don't have the permissions to add a runner. Here are the docs: https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/adding-self-hosted-runners .

Basically, if you add a runner Github will give you a link to download the client together with a key. This key is important. Maybe you can send the link and the key to me over slack if you like the idea of Github Actions. Our experience with ExaPF, short as it maybe, has been very nice so far. 

Edit: A runner is added under "Settings -> Actions -> Self-hosted Runners"


sshin23:
Thanks for putting this together. Merging this to master

PR #5, opened 2020-10-08
Replace Parameters.@with_kw by Base.@kwdef

frapac:
That allows to remove `Parameters.jl`dependency


sshin23:
Didn't know about this! Thank you @frapac

PR #6, opened 2020-11-23
No-MKL support

sshin23:
With this change, if MadNLP is built with OpenBLAS, MadNLP does not require MKL at all. @michel2323 could you check MadNLP builds properly on Summit?


michel2323 (approved):
Works on Summit now.

PR #7, opened 2020-12-08
Allow for CUDA 2.x

michel2323:
ProxAL/ExaPF require CUDA.jl 2.x which breaks the dependencies since MadNLP does not allow CUDA.jl 2.x. This should fix that.


sshin23:
Merged. Thanks, @michel2323!

PR #12, opened 2020-12-17
Replacing artifacts by jlls

sshin23:
Now jlls are used instead of artifacts for MUMPS, OpenBLAS, MKL. This allows building MUMPS without compilation. Now we can use MadNLP on windows.

@michel2323 Could you check if this works on powerpc? You may need to `ENV["MADNLP_BLAS"]="openblas"` before `build MadNLP`.


michel2323:
It does work on PPC with the option you mentioned. Otherwise it fails as MKL is not available. Maybe one could change the default for `Sys.ARCH == :powerpc64le`?

deps/build.jl
```diff
@@ -1,22 +1,14 @@
 # MadNLP.jl
 # Created by Sungho Shin (sungho.shin@wisc.edu)
 
-
-Sys.iswindows() && error("Windows is currently not supported.")
+blasvendor=(haskey(ENV,"MADNLP_BLAS") && ENV["MADNLP_BLAS"]=="openblas") ? :openblas : :mkl
```
michel2323, line 4: I guess this should be `:openblas` for anything other than `Sys.ARCH ==  :x86_64`

sshin23:
Now BLAS vendor is set as MKL if `MKL_jll.best_platform != nothing` and OpenBLAS otherwise.

sshin23:
Merged to master. Thanks for the help @michel2323!

PR #23, opened 2021-03-09
Removing MPI and CUDA dependency

sshin23:
With this change, MPI is not required anymore.

CUDA is now loaded lazily via Require.jl. To use GPU solvers, now the user needs to import CUDA to the main module first. 

@frapac Finally realized that Require.jl was gold. Let me know if this may cause any issues on your end.


README.md
```diff
@@ -88,6 +88,19 @@ model = CUTEstModel("PRIMALC1")
...
 
+### Using special linear solvers
+In order to use GPU solvers, `CUDA` should be imported to the `Main` module.
+```julia
+using MadNLP, CUDA
+model = Model(()->MadNLP.Optimizer(linear_solver=MadNLP.LapackGPU))
+# ...
+```
+
+In order to use multi-threaded solvers (`Schur` and `Schwawrz`), julia sesseion should be started with `-t` flag.
```
frapac, line 99: I think there is a little typo: `Julia session`?

frapac: and `Schwarz`?

src/LinearSolvers/mumps.jl
```diff
@@ -4,13 +4,14 @@
...
 import ..MadNLP:
-    parsefile, SVector, setindex, MPI,
+    parsefile, SVector, setindex, dlopen,
     @kwdef, Logger, @debug, @warn, @error,
     SparseMatrixCSC, SubVector, StrideOneVector, MUMPS_seq_jll,
     SymbolicException,FactorizationException,SolveException,InertiaException,
     AbstractOptions, AbstractLinearSolver, set_options!,
     introduce, factorize!, solve!, improve!, is_inertia, inertia, findIJ, nnz
 
+# const libmpiseq = joinpath(MUMPS_seq_jll.artifact_dir,"lib","libmpiseq.so")
```
frapac, line 14: Maybe remove dead comment?

frapac (commented):
I think getting rid of `CUDA` as a direct dependency is a good thing. :+1: 

However, I don't know if `Requires.jl` is the good solution for that. See e.g. the discussion here: 
https://github.com/JuliaGPU/KernelAbstractions.jl/pull/200#issuecomment-759275424

The drawbacks being: 
- difficulties in the precompilation (especially as `Requires.jl` induces cache invalidation)
- impracticable compilation with `PackageCompiler`. 

They have chosen a different approach in `KernelAbstractions`, by moving all deps to CUDA in a special `lib/CUDAKernels` directory. But the way they are loading this external dependency is not smooth. 

Here, instead of using `Requires`, I would instead use the interface of `AbstractLinearSolver` you defined in `src/LinearSolvers/linearsolvers.jl`, which allows modularity. All linear solvers should follow this  interface. I would just implement the base linear solvers in `src/`, and implement more complicated linear solvers apart (maybe in a third-party `lib` directory). Then, we would chose which external linear solvers to load in the `deps/build.jl` script.

But this is a just a hot take. I don't know what is the correct solution for MadNLP at the end :)

sshin23:
@frapac Thanks for the all comments!

Yeah, I can definitely see some of the potential issues with Requires.jl. I was also considering making each linear solver an independent package, but that may confuse the user. But for solvers like Schur and Schwarz, I may make them independent, because most users will not need them.

For HSL, Pardiso, and CUDA, I think the approach you suggested is appropriate. I'll try to implement it soon.

And thanks for the review! 👍

PR #30, opened 2021-04-29
Update for Plasmo v0.4.

jalving:
Plasmo.jl v0.4 (the current master) has made some drastic changes under the hood, but almost all of the interface has stayed the same.  This PR makes some minor updates to the MadNLP-Plasmo interface to make it work for v0.3.3 and v0.4.  The only major change was updating how the underlying backend is queried from an optinode since MadNLP directly accesses attributes. 

I also updated the Project.toml for MUMPS_seq_jll since it was grabbing v5.4.0 which caused MadNLP to fail to compile.


sshin23:
Thanks @jalving! I'll look into the changes

sshin23 (approved):

PR #33, opened 2021-05-20
Optimize copy of a SparseMatrixCOO to a dense matrix

frapac:
When solving dense nonlinear optimization problem, copying a `SparseMatrixCOO` to a dense array amount to up to 11% of the total running time in `MadNLP.optimize!`. This PR should add an `@inbounds` macro to optimize the function `_copyto!`


frapac:
Tests are breaking because of IterativeSolvers. See PR https://github.com/sshin23/MadNLP.jl/pull/34

sshin23:
@frapac Thanks for working on this! Later I think later we should separately define a dense nonlinear program type and natively support dense NLPs. But for now, this looks good.

Issue #35, opened 2021-06-02
[CUDA] Add support for cuSOLVER's sytrs 


frapac:
At the moment, when using Bunch-Kaufman with `LapackGPU`, MadNLP is using cusolver to compute the factorization on the GPU, then copy the factorization back to the CPU to do the triangular solver with OpenBlas. 

The reason of this additional data transfer is that the routine `sytrs` is not implemented properly in cusolver (there exists a function `cusolverDnDsytrs` in the API, but apparently it doesn't do anything, and this function was not even referenced in the cusolver's documentation).

However, it looks like they have included a new `sytrs` function in CUDA 11.3.1: cf the [documentation](https://docs.nvidia.com/cuda/cusolver/index.html#cusolverDnXsytrs) (strangely, they don't mention this new function in the release note).
It would be interesting to test this new function inside MadNLP (it would require to wrap the new `cusolverDnXsytrs` inside `CUDA.jl`). If it works, we would save one expensive data transfer :-)

sshin23:
I just tested the new `cusolverDnXsytrs` in CUDA 11.3.1 and it works! Check [this commit](https://github.com/sshin23/MadNLP.jl/commit/fcc9b97e33326bd730487218e2e6c8eb2e92f74d) in the ss/cuda_sytrs branch. It should work with `CUDA#master` and `GPUCompiler#master`. 

In this commit, I'm directly ccall-ing the cusolverDnXsytrs, as CUDA.jl doesn't have a wrapper for this function yet. A bit weird thing is that cusolverDnXsytrs doesn't work with Cint but works with Int64. So I needed to create another ipiv array for an Int64 array.

I'll work on a PR for this change once the current CUDA#master is released in their next version.

sshin23:
And in order to check the inertia without moving the LDLT factorization result to CPU, `inertia(M::LapackGPU.Solver)` needs to be implemented. [This part](https://github.com/sshin23/MadNLP.jl/blob/2ce1a302b38345beccd6905051659721c9063a4c/src/LinearSolvers/lapackcpu.jl#L168-L189) should be modified.

frapac:
That's good to hear, I am looking forward to test that! 
About the inertia, do you know by any chance if we could use MadNLP in inertia-free mode with LapackGPU? And if so, how would it impact the performance?

sshin23:
[...]

PR #38, opened 2021-06-15
Add support for cuSOLVER's sytrs

sshin23:
This PR is a follow-up of #35. With this change, LapackGPU now uses cuSOLVER's sytrs, instead of the LapackCPU's sytrs.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,7 +31,7 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 [compat]
 AmplNLReader = "~0.9"
 BinaryProvider = "~0.5"
-CUDA = "~1.3, ~2, 3.0"
+CUDA = "~3.3"
 IterativeSolvers = "~0.9.1"
 JuMP = "~0.21"
 LightGraphs = "~1.3"
--- a/Project.toml
+++ b/Project.toml
@@ -44,7 +44,7 @@ OpenBLAS32_jll = "~0.3"
 Plasmo = "~0.3"
 SolverCore = "~0.1"
 StaticArrays = "~0.12, 1.0"
-julia = "1.5"
+julia = "1.6"
 
 [extras]
 AmplNLReader = "77dd3d4c-cb1d-5e09-9340-85030ff7ba66"
--- a/Project.toml
+++ b/Project.toml
@@ -31,7 +31,7 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 [compat]
 AmplNLReader = "~0.9"
 BinaryProvider = "~0.5"
-CUDA = "~3.3"
+CUDA = "~3"
 IterativeSolvers = "~0.9.1"
 JuMP = "~0.21"
 LightGraphs = "~1.3"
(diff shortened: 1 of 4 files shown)


sshin23:
It seems that the issue is caused by moonshot's using its local installation of CUDA Toolkit instead of the standard artifact (see [here](https://github.com/sshin23/MadNLP.jl/runs/2825722319?check_suite_focus=true#step:5:170)). 

@frapac Is there a possibility that the use of artifact disallowed in moonshot?

Project.toml
```diff
@@ -31,7 +31,7 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 [compat]
 AmplNLReader = "~0.9"
 BinaryProvider = "~0.5"
-CUDA = "~1.3, ~2, 3.0"
+CUDA = "~3.3"
```
frapac, line 34: I am wondering if that bound is too restrictive? Do we need `CUDA.jl` v3.3 to use CUDA 11.3.1?

src/LinearSolvers/lapackgpu.jl
```diff
@@ -2,10 +2,17 @@ module LapackGPU
 
 import ..MadNLP:
     @kwdef, Logger, @debug, @warn, @error,
-    CUBLAS, CUSOLVER, CuVector, CuMatrix,
+    CUBLAS, CUSOLVER, CuVector, CuMatrix, R_64F,
```
frapac, line 5: Out of curiosity, what does `R_64F` stands for?

src/LinearSolvers/lapackgpu.jl
```diff
@@ -65,52 +76,54 @@ function solve!(M::Solver,x)
...
         Int32(size(M.fact,1)),M.fact,Int32(size(M.fact,2)),
         M.etc[:ipiv],M.work,M.lwork[],M.info)
 
-    # need to send the factorization back to cpu to call mkl sytrs --------------
-    haskey(M.etc,:fact_cpu) || (M.etc[:fact_cpu] = Matrix{Float64}(undef,size(M.dense)))
-    haskey(M.etc,:ipiv_cpu) || (M.etc[:ipiv_cpu] = Vector{Int}(undef,size(M.dense,1)))
-    copyto!(M.etc[:fact_cpu],M.fact)
-    copyto!(M.etc[:ipiv_cpu],M.etc[:ipiv])
-    # ---------------------------------------------------------------------------
+    # # need to send the factorization back to cpu to call mkl sytrs --------------
```
frapac, line 96: Maybe remove the dead comment?

frapac (commented):
Very nice work! That will help a lot in our use case. 
The only thing I am not sure with: do we need to restrict the bound for `CUDA.jl` here?

Project.toml
frapac: I am wondering whether, instead of forcing the CUDA version to be `>= 3.3`, we could instead add a `if` statement in the code to load the function `sytrs` only if `CUDA 11.3` is detected on the system. By doing that, we would avoid a breaking change.

sshin23: That might be a better idea 😊

sshin23:
Merged to master

PR #45, opened 2021-06-24
Add benchmarks

sshin23:
This PR aims to build a comprehensive benchmark suite for MadNLP.

Some of the candidate benchmark problem sets are:
- CUTEst problem set via [CUTEst.jl](https://github.com/JuliaSmoothOptimizers/CUTEst.jl)
- The OPF problems in [pglib-opf](https://github.com/power-grid-lib/pglib-opf) via [PowerModels.jl](https://github.com/lanl-ansi/PowerModels.jl)

The goal is to allow automatic comparison of two particular commits (or branches).


benchmark/.Project.toml
```diff
@@ -0,0 +1,8 @@
+[deps]
```
frapac, line 1: Why do we have two Project.toml here? (`.Project.toml` and `Project.toml`)

benchmark/benchmark-cutest.jl
```diff
@@ -0,0 +1,79 @@
...
+    @everywhere solver = nlp -> madnlp(nlp,linear_solver=MadNLPMa57,max_wall_time=900.,tol=1e-6)
+    @everywhere using MadNLP, MadNLPHSL
+elseif SOLVER == "current"
+    @everywhere solver = nlp -> madnlp(nlp,linear_solver=MadNLPMa57,max_wall_time=900.,tol=1e-6)
+    @everywhere using MadNLP, MadNLPHSL
+elseif SOLVER == "ipopt"
+    @everywhere solver = nlp -> ipopt(nlp,linear_solver="ma57",max_cpu_time=900.,tol=1e-6)
+    @everywhere using NLPModelsIpopt
+elseif SOLVER == "knitro"
+    # TODO
```
frapac, line 25: If you don't have a Knitro's license, I think that something we could discuss with Artelys

benchmark/benchmark-cutest.jl
```diff
@@ -0,0 +1,79 @@
...
+    broadcast(decodemodel,probs)
+    
+    println("Solving problems")
+    retvals = pmap(prob->evalmodel(prob,solver),probs)
+    time   = [retval.elapsed_time for retval in retvals]
+    status = [get_status(retval.status) for retval in retvals]
+    time,status
+end
+
+exclude = [
```
frapac, line 62: Maybe add a comment to explain why we are excluding these tests?

benchmark/benchmark-power.jl
```diff
@@ -0,0 +1,76 @@
+const PGLIB_PATH = ENV["PGLIB_PATH"]
+
+include("config.jl")
+
+using MathOptInterface; const MOI=MathOptInterface
```
frapac, line 5: I would maybe add a new line instead of a semi-colon `;`

benchmark/benchmark-cutest.jl
```diff
@@ -0,0 +1,79 @@
...
+
+
+probs = CUTEst.select()
+
+println(probs)
+filter!(e->!(e in exclude),probs)
+
+time,status = benchmark(solver,probs;warm_up_probs = ["EIGMINA"])
+
+save_object("name-cutest.jld2",probs)
```
frapac, line 77: Something to be careful with JLD: the format is not 100% specified, so I am not sure we could read the results in 2 years for instance (I can't read the JLD I exported during my PhD)

frapac: My rule of thumb is exporting as much as I can flat files, in text format. My personal opinion is that this kind of format should be reserved to large dataset.

benchmark/config.jl
```diff
@@ -0,0 +1,21 @@
+using Pkg, Distributed, JLD2
+
+const NP = ARGS[1]
+const SOLVER = ARGS[2]
+
+addprocs(parse(Int,NP),exeflags="--project=.")
```
frapac, line 6: Do we need to add explicitly the flag `--project` when using `Distributed.jl`?

frapac (commented):
Looks good to me! It's nice to generate benchmarks in a distributed way. 

For our local benchmarks, we are trying to define a `Makefile` for each benchmark, so we could initiate the benchmark and run it simply as 
```sh
make benchmark
```
(see e.g. [this file](https://github.com/exanauts/BatchHessianArtifact/blob/master/Makefile))
But it would be specific to Unix system, though.

benchmark/config.jl
sshin23: Yes, according to the [Distributed.jl documentation](https://docs.julialang.org/en/v1/manual/distributed-computing/)

benchmark/benchmark-cutest.jl
sshin23: Gotcha, we can export it as csv

benchmark/.Project.toml
sshin23: Thanks for catching this; Project.toml should be removed.

We separately have `.Project.toml` because `Pkg.add("MadNLP")` in the script adds MadNLP entry to the `Project.toml` file. So, locally running the benchmark and then pushing it to repo can cause an unwanted change in the Project.toml file.

frapac: I see, it makes sense indeed!

benchmark/benchmark-cutest.jl
frapac: I think CSV would be better, indeed. And we would be able to import the results in an other language than Julia :-)

PR #46, opened 2021-06-25
[DNMY] WIP: Add support to MOI 0.10

frapac:
Fix breaking changes for MOI 0.10. 
This is still a work-in-progress, as this PR is using `MOI.DeprecatedTest` for the testing. It remains to port to the new `MOI.Test` module once it will be fully merged inside MOI.


frapac:
Tests are not passing yet: it remains to port the MOI tests to the new `MOI.Tests`.

PR #58, opened 2021-08-20
Add AbstractKKTSystem structure

frapac:
First step towards support of dense NLP 

* implement SparseReducedKKTSystem and SparseAugmentedKKTSystem
* move `Options` in a new file `src/options.jl` 
* `Solver`: move all KKT structures in a new field `kkt`
* `Solver`: remove closures and implement everything in proper functions 
* `Solver`: callbacks are now implemented as proper Julia function (better for compiler, and allow to use multiple dispatch in the future)


src/kktsystem.jl
```diff
@@ -0,0 +1,339 @@
+
+abstract type AbstractKKTSystem{T, MT} end
```
frapac, line 2: I am not 100% sure about that name. I am wondering whether `AbstractKKTLinearSystem` would be more explicit?

src/kktsystem.jl
```diff
@@ -0,0 +1,339 @@
...
+        ind_ineq, ind_fixed, ind_aug_fixed, jac_scaling,
+    )
+end
+
+
+#=
+    SparseAugmentedKKTSystem
+=#
+
+struct SparseAugmentedKKTSystem{T, MT} <: AbstractSparseKKTSystem{T, MT}
```
frapac, line 237: I split the `SparseAugmentedKKTSystem` from `SparseReducedKKTSystem` for clarity purpose. Let me know if you think it would be better to merge these two together

src/matrixtools.jl
```diff
@@ -45,42 +55,62 @@ end
...
+    return map
+end
+
 function get_coo_to_dense(coo::SparseMatrixCOO{Tv,Ti}) where {Tv,Ti<:Integer}
     dense = Matrix{Float64}(undef,coo.m,coo.n)
     return dense, ()->copyto!(dense,coo)
 end
 
-copyto!(dense::Matrix{Tv},coo::SparseMatrixCOO{Tv,Ti}) where {Tv,Ti<:Integer} = _copyto!(dense,coo.I,coo.J,coo.V)
+function Matrix{Tv}(coo::SparseMatrixCOO{Tv,Ti}) where {Tv,Ti<:Integer}
```
frapac, line 109: I add a proper constructor to build a dense matrix from a `SparseMatrixCOO`

src/matrixtools.jl
```diff
@@ -45,42 +55,62 @@ end
...
-    # elseif mtype == :cucsc
-    #     get_coo_to_com = get_coo_to_cucsc
     elseif mtype == :dense
         get_coo_to_com = get_coo_to_dense
-    # elseif mtype == :cudense
-    #     get_coo_to_com = get_coo_to_cudense
     end
 end
 
+function SparseMatrixCSC{Tv, Ti}(coo::SparseMatrixCOO{Tv,Ti}) where {Tv,Ti <: Integer}
```
frapac, line 63: This overload `SparseMatrixCSC`'s constructor to build a CSC directly from a COO

src/options.jl
```diff
@@ -0,0 +1,96 @@
...
+    max_wall_time::Float64 = 1e6
+    s_max::Float64 = 100.
+
+    # NLP options
+    kappa_d::Float64 = 1e-5
+    fixed_variable_treatment::FixedVariableTreatments = MAKE_PARAMETER
+    jacobian_constant::Bool = false
+    hessian_constant::Bool = false
+    reduced_system::Bool = true
+    kkt_system::KKTLinearSystem = SPARSE_REDUCED_KKT_SYSTEM
```
frapac, line 30: We now specify which `kkt_system` type we want with this option

sshin23 (changes requested):
Hey @frapac, the PR looks awesome! Indeed AbstractKKTSystem can significantly enhance the modularity.

My only concern is the treatment for the "unreduced" system. I think your approach of treating it as an alternative KKT system formulation is the correct approach. But formulating it as a block 3x3 unreduced system requires some extra treatment. So, I'd suggest the following:

- Remove `reduced_system` option and let this be treated entirely by setting `kkt_system` option.
- We can make anything that ends with `_reduced!` (e.g., `set_aug_reduced!`) the default method.
- Anything that ends with `_unreduced!` can be made as a special treatment for the unreduced system. For example, we can make
`set_aug_unreduced!(...)` into `set_aug_reduced!(::SparseUnreducedKKTSystem, ...)`

src/kktsystem.jl
frapac: Good catch!

sshin23:
🎉 [benchmark cutest master,current,ipopt --quick --gcoff](https://github.com/sshin23/MadNLP.jl/files/7070210/time-cutest.3.pdf)

src/interiorpointsolver.jl
```diff
@@ -265,9 +265,8 @@ function eval_grad_f_wrapper!(ipp::Solver, f::Vector{Float64},x::Vector{Float64}
     nlp = ipp.nlp
     cnt = ipp.cnt
     @trace(ipp.logger,"Evaluating objective gradient.")
-    ∇f = view(f, 1:nlp.n)
-    cnt.eval_function_time += @elapsed nlp.obj_grad!(∇f,view(x,1:nlp.n))
-    ∇f.*=ipp.obj_scale[]
+    cnt.eval_function_time += @elapsed nlp.obj_grad!(f,view(x,1:nlp.n))
+    f.*=ipp.obj_scale[]
     cnt.obj_grad_cnt+=1
```
sshin23, line 270: The issue was here; for some reason, sending `∇f` caused some extra allocations. Maybe the reason comes from CUTEst.jl. If the gradient callback checks the dimension of `f`, the current change can be problematic

sshin23: For a similar reason, if we replace `view(x,1:nlp.n)` with `x`, allocations are significantly reduced. But if the gradient callback checks the input size, it can cause issue.

sshin23: Indeed, the issue was here:
https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/main/src/julia_interface.jl#L105-L110

In order to make it C function compatible, CUTEst.jl internally converts the inputs to `Vector{Float64}`. For "EIGMINA" in CUTEst, after allowing StrideOneVector{Float64} in the inputs, the allocation dramatically reduces:
from 1.213 KiB to 759.922 KiB. 

I'll create a PR for CUTEst.jl. I think the benchmark result will look a lot better after this...

frapac: That's very interesting. Good catch for the issue with CUTest.jl. I think indeed there is no need to create a new array if the input `x` is one-strided ... I did this change because of the custom wrapper we developed for ExaPF: here, we need a gradient with the correct dimension (as we are using `mul!` operation afterwards to compute the adjoint). I don't know what's the default: should we pass to the callbacks all arrays with the correct dimension, or should the callbacks do all the checks internally?

sshin23: I think the latter would be ideal, but the performance of MadNLP shouldn't degrade too much even if the callback requires the correct dimension. I added a "buffered" option for NLPModels.jl interface so that there is a buffer vector that exactly matches the input dimensions for the callbacks. NLPModels.jl interface handles copying to/from the buffer vector. But if the user knows that there won'te be a problem, the user can turn this off.

I made a PR for CUTEst.jl (https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/269) to allow view inputs. If this gets merged, we can turn off buffered option for the benchmark.

sshin23:
Benchmark results:
[time](https://github.com/sshin23/MadNLP.jl/files/7072937/time-cutest.4.pdf)
[memory](https://github.com/sshin23/MadNLP.jl/files/7072938/memory-cutest.3.pdf)

frapac:
Thanks for the help in merging this PR, @sshin23 !

PR #59, opened 2021-08-24
Deactive logs in benchmark scripts

frapac:
+ some minor formatting

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,5 +1,10 @@
 To run benchmarks, run the following script.
 ```
 # to compare current version and master using 10 cores and CUTEst and PowerModels.jl problems,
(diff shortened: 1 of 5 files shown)


sshin23 (commented):
Thanks, @frapac for the improvements! I was thinking that the user may want to see the progress of running the benchmark. Now that we don't print the solver output, maybe we can at least show which problem it is currently solving?

sshin23:
other than that, everything looks good 👍

frapac:
Good point. I think we already have some information with: https://github.com/sshin23/MadNLP.jl/blob/master/benchmark/benchmark-power.jl#L42
But you are right: would you like me to add an option to active solver's log as well?

sshin23:
Aha, yes you're right. And yes, I think it would be good to make it as a command line option.

PR #64, opened 2021-08-30
performance improvement

sshin23:
In this PR, we make several changes for performance improvement.

-   drop the support for unreduced KKT system
-   bugfix in solve_refine! (reduces allocation)
-  use @simd macro when possible
-   removed specialization functions when not performance-critical
-  use apply_step! function

### DIFF
--- a/src/interiorpointsolver.jl
+++ b/src/interiorpointsolver.jl
@@ -92,20 +92,14 @@ mutable struct Solver{KKTSystem} <: AbstractInteriorPointSolver
     p::Vector{Float64}
     px::StrideOneVector{Float64}
     pl::StrideOneVector{Float64}
-    pzl::Union{Nothing,StrideOneVector{Float64}}
-    pzu::Union{Nothing,StrideOneVector{Float64}}
 
     _w1::Vector{Float64}
     _w1x::StrideOneVector{Float64}
     _w1l::StrideOneVector{Float64}
-    _w1zl::Union{Nothing,StrideOneVector{Float64}}
-    _w1zu::Union{Nothing,StrideOneVector{Float64}}
 
     _w2::Vector{Float64}
     _w2x::StrideOneVector{Float64}
     _w2l::StrideOneVector{Float64}
-    _w2zl::Union{Nothing,StrideOneVector{Float64}}
-    _w2zu::Union{Nothing,StrideOneVector{Float64}}
 
     _w3::Vector{Float64}
     _w3x::StrideOneVector{Float64}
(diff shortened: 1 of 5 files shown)


src/interiorpointsolver.jl
```diff
@@ -709,9 +690,8 @@ function regular!(ips::AbstractInteriorPointSolver)
         switching_condition = is_switching(varphi_d,ips.alpha,ips.opt.s_phi,ips.opt.delta,2.,ips.opt.s_theta)
         armijo_condition = false
         while true
-            ips.x_trial .= ips.x .+ ips.alpha.*ips.dx
+            apply_step!(ips.x_trial,ips.x,ips.dx,ips.alpha)
```
frapac, line 693: instead of using a new function `apply_step!`, how about using directly a BLAS function? 
```julia
copyto!(ips.x_trial, ips.x)
axpy!(ips.alpha, ips.dx, ips.x_trial)
```

src/interiorpointsolver.jl
```diff
@@ -751,9 +731,9 @@ function regular!(ips::AbstractInteriorPointSolver)
         adjusted > 0 &&
             @warn(ips.logger,"In iteration $(ips.cnt.k), $adjusted Slack too small, adjusting variable bound")
 
-        ips.l.+=ips.alpha.*ips.dl
-        ips.zl_r.+=ips.alpha_z.*ips.dzl
-        ips.zu_r.+=ips.alpha_z.*ips.dzu
+        apply_step!(ips.l,ips.l,ips.dl,ips.alpha)
```
frapac, line 734: Same here? 
```julia
axpy!(ips.alpha, ips.dl, ips.l)
```

src/interiorpointsolver.jl
```diff
@@ -1089,65 +1068,29 @@ function second_order_correction(ips::AbstractInteriorPointSolver,alpha_max::Flo
...
-    u_diag  .= x_ur .- xu_r
-end
 function set_aug_diagonal!(kkt::AbstractKKTSystem, ips::Solver)
-    _set_aug_diagonal_reduced!(kkt.pr_diag, kkt.du_diag, ips.x, ips.xl, ips.xu, ips.zl, ips.zu)
-end
-function set_aug_diagonal!(kkt::SparseUnreducedKKTSystem, ips::Solver)
-    _set_aug_diagonal_unreduced!(kkt.pr_diag, kkt.du_diag, kkt.l_lower, kkt.u_lower, kkt.l_diag, kkt.u_diag,
-        ips.zl_r, ips.zu_r, ips.xl_r, ips.xu_r, ips.x_lr, ips.x_ur,
-    )
+    kkt.pr_diag .= ips.zl./(ips.x.-ips.xl) .+ ips.zu./(ips.xu.-ips.x)
```
frapac, line 1080: Nice that we can get rid of the auxiliary function!

src/interiorpointsolver.jl
```diff
@@ -1224,179 +1139,179 @@ is_valid(args...) = all(is_valid(arg) for arg in args)
 
 function get_varphi(obj_val,x_lr,xl_r,xu_r,x_ur,mu)
     varphi = obj_val
-    for i=1:length(x_lr)
-        xll = x_lr[i]-xl_r[i]
-        @inbounds xll < 0 && return Inf
-        @inbounds varphi -= mu*log(xll)
+    @simd for i=1:length(x_lr)
```
frapac, line 1142: Do we observe a difference when using `@simd`?

frapac (commented):
It overall looks good to me. I am just not sure to understand why we get rid of the `SparseUnreducedKKTSystem` here? Was it really performance detrimental? 

About the `@simd` macro, I am wondering if in the long term we could incorporate a proper package for SIMD vectorization in MadNLP. E.g. 
- [LoopVectorizations.jl](https://github.com/JuliaSIMD/LoopVectorization.jl)
- [Tullio.jl](https://github.com/mcabbott/Tullio.jl)

src/interiorpointsolver.jl
sshin23: Yes, using BLAS makes more sense :+1:

src/interiorpointsolver.jl
sshin23: not quite sure, I think that will need a separate benchmark

frapac (approved):
This PR looks good to me now. Nice that we can remove the auxiliary functions we implemented in #58 ! 
I will start working on implementing the KKT right-hand-side structure once this PR merged in master.

Issue #67, opened 2021-09-02
Retire `NonlinearProgram` in favor of `NLPModels`


sshin23:
Currently, we have `NonlinearProgram`, which provides basic model information to MadNLP. But all of the same functionality can be provided by `NLPModels.jl`, and it even has more features. In the future, `NonlinearProgram` will be replaced by `NLPModels.jl `, and the current MOI interface will be replaced by a thin wrapper of `NLPModelsJuMP`.

PR #69, opened 2021-09-04
Retire NonlinearProgram

sshin23:
This PR implements the suggestions in #67


src/Interfaces/MOI_interface.jl
```diff
@@ -874,25 +872,23 @@ const status_dual_dict = Dict(
...
     scale = (model.sense == MOI.MAX_SENSE) ? -1 : 1
-    return scale * model.nlp.obj_val
+    return scale * model.ips.obj_val
 end
 
 function MOI.get(model::Optimizer, attr::MOI.VariablePrimal, vi::MOI.VariableIndex)
     MOI.check_result_index_bounds(model, attr)
     check_inbounds(model, vi)
-    return model.nlp.x[vi.value]
+    return model.ips.x[vi.value]
```
frapac, line 891: I think this is a net improvement: instead of having two primal variables `ips.x` and `ips.nlp.x`, now everything is stored inside `ips`

src/Interfaces/MOI_interface.jl
```diff
@@ -1101,58 +1083,75 @@ end
...
 num_variables(model::Optimizer) = length(model.variable_info)
 
-function get_obj_scale(sense)
-    sense == MOI.MIN_SENSE && return 1.
-    sense == MOI.MAX_SENSE && return -1.
-    return 0.
-end
-
-function NonlinearProgram(model::Optimizer)
+struct MadNLPModel <: AbstractNLPModel{Float64,Vector{Float64}}
```
frapac, line 1086: How about parameterizing the type? 
```julia
struct MadNLPModel{T} <: AbstractNLPModel{T,Vector{T}}
```

src/Interfaces/MOI_interface.jl
```diff
@@ -1101,58 +1083,75 @@ end
...
 function MOI.optimize!(model::Optimizer)
-    model.nlp = NonlinearProgram(model)
-    model.ips = Solver(model.nlp;option_dict=copy(model.option_dict))
-    optimize!(model.ips)
+    nlp = MadNLPModel(model)
+    ips = Solver(nlp;option_dict=copy(model.option_dict))
+    model.ips = ips
+    optimize!(ips)
+
+    ips.obj_val = ips.obj_val/ips.obj_scale[]
```
frapac, line 1150: I would suggest to move  the lines 1150-1153 in a separate function

src/interiorpointsolver.jl
```diff
@@ -292,7 +292,7 @@ function eval_jac_wrapper!(ipp::Solver, kkt::AbstractKKTSystem, x::Vector{Float6
     ns = length(ipp.ind_ineq)
     @trace(ipp.logger, "Evaluating constraint Jacobian.")
     jac = get_jacobian(kkt)
-    cnt.eval_function_time += @elapsed nlp.con_jac!(jac, view(x,1:nlp.n))
+    cnt.eval_function_time += @elapsed jac_coord!(nlp,view(x,1:get_nvar(nlp)),jac)
```
frapac, line 295: I don't know how this will play with `DenseKKTSystem`

src/interiorpointsolver.jl
```diff
@@ -362,15 +383,15 @@ function Solver(nlp::NonlinearProgram;
         DenseKKTSystem{Float64, VT, MT}(nlp, ind_cons)
     end
 
-    xl = [nlp.xl;view(nlp.gl,ind_cons.ind_ineq)]
-    xu = [nlp.xu;view(nlp.gu,ind_cons.ind_ineq)]
-    x = [nlp.x;zeros(ns)]
-    l = nlp.l
-    zl= [nlp.zl;zeros(ns)]
-    zu= [nlp.zu;zeros(ns)]
+    xl = [get_lvar(nlp);view(get_lcon(nlp),ind_cons.ind_ineq)]
```
frapac, line 386: Here I would maybe add a few type hinting for the compiler. Like: 
```julia
xl_lvar = get_lvar(nlp)::Vector{Float64}
xl_lcon = view(get_lcon(nlp)::Vector{Float64},ind_cons.ind_ineq)
xl =Float64[xl_lvar; xl_lcon]
```
(but might be a bit ugly)

test/madnlp_dense.jl
```diff
@@ -1,19 +1,79 @@
 using Test
+import MadNLP: jac_structure!, hess_structure!, obj, grad!, cons!, jac_coord!, hess_coord!, jac_dense!, hess_dense!
+using NLPModels
 using LinearAlgebra
 using SparseArrays
 using Random
 
+struct DenseQP <: AbstractNLPModel{Float64,Vector{Float64}}
```
frapac, line 8: That structure is definitely cleaner, and by itself justifies your whole PR!!

frapac: How about renaming it as `DenseDummyQP`? Just to convey the idea that the data are generated randomly and the problem is not very interesting, outside the scope of this tests suite?

test/madnlp_dense.jl
```diff
@@ -1,19 +1,79 @@
...
+    meta::NLPModels.NLPModelMeta{Float64, Vector{Float64}}
+    P::Matrix{Float64} # primal hessi
```
frapac, line 10: I think there is a typo in the comment? `primal hessian`

test/madnlp_dense.jl
```diff
@@ -1,19 +1,79 @@
...
+    A::Matrix{Float64} # constratin jacobian
```
frapac, line 11: Same here? `constraints Jacobian`

test/madnlp_dense.jl
```diff
@@ -1,19 +1,79 @@
...
+        index += 1
+    end
+end
+# Hessian: dense callback
+function hess_dense!(qp::DenseQP, x, l,hess::AbstractMatrix; obj_weight=1.)
+    hess .= obj_weight .* qp.P
+end
+
 
 function build_qp_test(; n=100, m=10, fixed_variables=Int[])
```
frapac, line 65: I would maybe define this function directly as a constructor for `DenseQP`?
```julia
function DenseQP(; n=100, m=10, fixed_variables=Int[])
```

test/madnlp_dense.jl
```diff
@@ -181,12 +182,10 @@ function _compare_dense_with_sparse(n, m, ind_fixed)
     )
 
     nlp = build_qp_test(; n=n, m=m, fixed_variables=ind_fixed)
-    x, l = copy(nlp.x), copy(nlp.l)
-
     ips = MadNLP.Solver(nlp, option_dict=sparse_options)
     MadNLP.optimize!(ips)
 
-    # Reinit NonlinearProgram to avoid side effect
+    # Reinit DenseQP to avoid side effect
```
frapac, line 188: Here the issue was that `nlp.x`, `nlp.l` ... were updated at the end of the solve. Hence, restarting the resolution from the same `NonlinearProgram` wamstarted implicitly the solver. Maybe this is not the case anymore (in my opinion, the model should be immutable)

test/runtests.jl
```diff
@@ -12,9 +12,10 @@ import SparseArrays: sparse
         include("MOI_interface_test.jl")
     end
 
-    @testset "NLPModels interface" begin
-        include("nlpmodels_test.jl")
-    end
+    # this is temporarily commented out due to the incompatibility between NLPModels v0.17.2 and AmplNLReader v0.11.0
```
frapac, line 15: Ouch. I agree that this is slightly annoying

frapac (commented):
Very good job! I think this PR is a net improvement to MadNLP!

sshin23:
Hey @frapac, thanks for the review. Just a heads-up, there are a few more things that need to be done in this PR, and I think it will take some time to finish. One thing to discuss is the interface between hessian/jacobian callbacks and NLPModel. Currently, NLPModels have `jac_coord!` and `hess_coord`, whose names indicate that it should evaluate the _coordinates_ of jacobian and hessian. I think we should talk to JSO guys to define proper dense callback APIs. For now, I have defined `jac_dense!` and `hess_dense!` in MadNLP.

frapac:
I agree it would be a good idea to open an issue in  `NLPModels.jl` about that. I think we are not the only one willing to evaluate dense Jacobian/Hessian. 
Let me know if I could help in this!

sshin23:
@frapac Feel free to go ahead and open an issue in NLPModels.jl. You must have a better idea on this since you've been working on the dense stuff for a long time 😉

src/Interfaces/MOI_interface.jl
sshin23: now we have a separate result in `AbstractExecutionStats` of `SolverCore.jl`, but we don't allocate memory (just return it as a view).

src/Interfaces/MOI_interface.jl
sshin23: now we have a separate function `unscale!` and call it at the end of `optimize!`.

src/interiorpointsolver.jl
sshin23: For dense, we have a different callback function `jac_dense!`, but this can be changed after coordinating with JSO folks

src/interiorpointsolver.jl
sshin23: Here, we're not certain that `get_lvar(nlp)` is `Vector{Float64}`, since we're taking `AbstractNLPModel` as an input. One possibility is to write a function e.g., `function get_initial_vectors` and do specialization for each NLPModel type. But I don't think this would be that much performance-critical

PR #77, opened 2021-09-13
disabling rescale bound duals after penalty paramter update

sshin23:
In this PR, the rescaling of bound dual variables after  is disabled. This reduces the number of iterations for many instances: [iterations-cutest(7).pdf](https://github.com/sshin23/MadNLP.jl/files/7154891/iterations-cutest.7.pdf)


sshin23:
hmm... interestingly this makes several instances in MINLPTests to fail. I'll investigate what's causing the issue

PR #88, opened 2021-09-16
kkt error reduction

sshin23:
Originally we had KKT error reduction phase (Section 3.3.2), but dropped the support some time in the past. But the KKT error reduction phase seems to be giving us some substantial performance gain [benchmark.pdf](https://github.com/sshin23/MadNLP.jl/files/7179746/time-cutest.15.pdf). So we bring it back in this PR


frapac (approved):
Nice!!

PR #126, opened 2021-12-17
Add DenseCondensedKKTSystem 

frapac:
Add new `DenseCondensedKKTSystem <: AbstractCondensedKKTSystem` abstraction. Works both on CPU and on CUDA GPU. 
The `DenseCondensedKKTSystem` allows to reduce drastically the size of the KKT system if the problem has many inequality constraints.

- `DenseKKTSystem` has been refactorized so the dense Jacobian has a size `(m, n)` (instead of `(m, n+ns)`)
- the option `CONDENSED_KKT_SYSTEM` loads a `DenseCondensedKKTSystem` under the hood (names can be improved, of course, I welcome any suggestion) 
- `MadNLPTests.DenseDummyQP` has been slightly updated to load more challenging QP problem (with inequality constraints activated at the solution). We now can mix both inequalities and equality constraints to test more thoroughly the different `KKTSystem` 
- this PR can be cleaned further once #116  is merged

### DIFF
--- a/src/KKT/dense.jl
+++ b/src/KKT/dense.jl
@@ -20,17 +20,86 @@ struct DenseKKTSystem{T, VT, MT} <: AbstractReducedKKTSystem{T, MT}
     # KKT system
     aug_com::MT
     # Info
+    n_ineq::Int
     ind_ineq::Vector{Int}
     ind_fixed::Vector{Int}
-    jacobian_scaling::VT
+    constraint_scaling::VT
     # Buffers
     etc::Dict{Symbol, Any}
 end
 
+"""
+    DenseCondensedKKTSystem{T, VT, MT} <: AbstractCondensedKKTSystem{T, MT}
+
+Implement [`AbstractCondensedKKTSystem`](@ref) with dense matrices.
+
+Requires a dense linear solver to factorize the associated KKT system (otherwise an error is returned).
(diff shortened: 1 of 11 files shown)


src/KKT/dense.jl
```diff
@@ -119,37 +190,184 @@ function build_kkt!(kkt::DenseKKTSystem{T, VT, MT}) where {T, VT, MT}
...
+    # Build J' * Σₛ * J
+    mul!(W, kkt.jac_ineq', kkt.jac_ineq)
+
+    _build_condensed_kkt_system!(
+        kkt.aug_com, kkt.hess, kkt.jac,
+        kkt.pr_diag, kkt.du_diag,
+        kkt.ind_eq, n, kkt.n_eq,
+    )
+    treat_fixed_variable!(kkt)
 end
```
sshin23, line 318: It would be interesting to examine directly regularizing the Schur complement.

sshin23 (approved):
Nice work, @frapac! Just have a few minor comments. Other than those, I think it's good to go!

frapac:
Thanks for the feedback Sungho! I addressed your comments in the last few commits, let me know what do you think of this new version :)

sshin23 (approved):
Approve the changes

PR #152, opened 2022-03-14
Add interface for custom scaling

frapac:
This PR exposes the scaling of the problem to the user with two functions:
- `scale_objective` 
- `scale_constraints!`

The idea is to dispatch on the `AbstractNLPModel` being employed. By doing so, the user can overload the two functions to chose a different scaling. For instance, the user can scale the objective by a factor `1e-3`with:
```julia
function scale_objective(nlp::MySpecialModel, grad::AbstractVector; max_gradient=1e-8)
     return 1e-3
end
``` 

This code was already employed to scale the reduced model in the paper.


sshin23 (approved):
Great idea @frapac!

Issue #153, opened 2022-03-23
MadNLP breaks Julia logging macros


hshose:
I'm using Julia 1.7 with updated MadNLP 0.4.0:
```
(@v1.7) pkg> status MadNLP
      Status `~/.julia/environments/v1.7/Project.toml`
  [2621e9c9] MadNLP v0.4.0
```

Running `@info "test" 1` in the REPL produces the following output (as expected):
```
julia> @info "TEST" 1
┌ Info: TEST
└   1 = 1
```

When I first `import MadNLP` and then `@info "test" 1`, I get the following error:
```
julia> import MadNLP
julia> @info "TEST" 1
ERROR: MethodError: no method matching get_level(::String)
Closest candidates are:
  get_level(::MadNLP.Logger) at ~/.julia/packages/MadNLP/OUcjZ/src/utils.jl:20
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/MadNLP/OUcjZ/src/utils.jl:34
```

MadNLP seems to overwrite the default logging macros, which is not ideal when using the solver in larger projects.

frapac:
Indeed, I confirm that this is a bug in MadNLP (I am facing the same issue). I will try to find a workaround, that issue is indeed annoying.

PR #154, opened 2022-03-23
Do not overload Logging macros

frapac:
Solve #153


sshin23:
@frapac I was working on making CI independent of MKL, so that we can use the fresh install of julia in every CI (#157). If it works out, I believe [c498a88](https://github.com/MadNLP/MadNLP.jl/pull/154/commits/c498a887a42a275f5d0c96d4bf8628c07310c009) is not necessary

frapac:
Ok, let me know if you want me to revert that commit!

PR #157, opened 2022-03-25
MadNLPPardisoMKL move to MadNLPPardiso.jl

sshin23:
With this change, the `MadNLPPardisoMKL` is moved to `MadNLPPardiso.jl` extension pacakge, and now `MadNLPPardiso.jl` relies on `MKL_jll`. This change allows for the self-hosted testing not to require MKL.

### DIFF
--- a/lib/MadNLPPardiso/src/MadNLPPardiso.jl
+++ b/lib/MadNLPPardiso/src/MadNLPPardiso.jl
@@ -6,17 +6,21 @@ const INPUT_MATRIX_TYPE = :csc
 
 import Libdl: dlopen, RTLD_DEEPBIND
 import MadNLP:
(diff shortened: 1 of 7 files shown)


lib/MadNLPPardiso/src/MadNLPPardiso.jl
```diff
@@ -6,17 +6,21 @@ const INPUT_MATRIX_TYPE = :csc
 
 import Libdl: dlopen, RTLD_DEEPBIND
 import MadNLP:
-    @kwdef, Logger, @debug, @warn, @error,
+    MadNLP, @kwdef, Logger, @debug, @warn, @error,
```
frapac, line 9: isn't it a counter pattern?

frapac (approved):
Looks good to me!

PR #164, opened 2022-04-25
callback bug fix

sshin23:
This PR fixes the bug in the callback wrapper. With the change, the vector length sanity check implemented in the callback side, such as
```
@lencheck nlp.meta.nvar x
@lencheck nlp.meta.nnzj vals
```
should pass.


frapac (approved):
Looks good to me!

PR #167, opened 2022-05-17
AbstractKKTRHS: second attempt 

frapac:
This is a new attempt at implementing the `AbstractKKTRHS` abstraction. Compared to the previous one, this PR uses a cleaner interface for the KKT right-hand-side, and update the implementation of iterative refinement to be more modular. 

In details, this PR
- adds `ReducedKKTVector` and `UnreducedKKTVector` structures 
- removes module `MadNLPRichardson` and replaces it with a proper structure `RichardsonIterator` 
- rewrites the iterative refinement algorithm to be more modular 
- add documentation for the linear solver's interface


frapac:
Thanks for the feedback @sshin23 ! What was the complaint about the variable name?

sshin23:
@frapac oops, sorry. I think `values` in `KKTVector` is not intuitive. How about `full`?

frapac:
That's a good suggestion! I will update the name for both the attribute and the getter

Project.toml
```diff
@@ -19,7 +21,6 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 [compat]
 MINLPTests = "~0.5"
 MadNLPTests = "~0.2"
-MathOptInterface = "1"
```
sshin23, line 22: Why is `MOI` compat removed?

sshin23 (approved):
🎉

Project.toml
sshin23: And `MadNLPGPU` and `MadNLPTests` don't need to be here, no?

PR #174, opened 2022-06-06
Remove StrideOneVector alias

frapac:
StrideOneVector lead to type inference issues in MadNLP. This PR replaces `StrideOneVector` by using explicit typing in the impacted structure (`SparseKKTSystem` and `SparseUnreducedKKTSystem`). This PR might improve the time to first solve in MadNLP.


lib/MadNLPHSL/src/ma27.jl
```diff
@@ -48,7 +48,7 @@ mutable struct Solver <: AbstractLinearSolver
     logger::Logger
 end
 
-ma27ad!(n::Cint,nz::Cint,I::StrideOneVector{Cint},J::StrideOneVector{Cint},
+ma27ad!(n::Cint,nz::Cint,I::AbstractVector{Cint},J::AbstractVector{Cint},
```
sshin23, line 51: I agree that `StrideOneVector` is not ideal, but the motivation for using it was to make sure that when calling linear solvers, we don't accidentally pass AbstractVectors with non-uniform memory stride (in particular, we want it to be always 1 byte). My concern with this change is that we don't have anything that prevents `ma27ad!` to accept a pointer to a non-stride-one-array. Of course, we're using these only internally within MadNLP, so shouldn't be a problem if we write the code carefully in the future, but it raises a bit of concern to me. Any thoughts on this @frapac?

frapac: To be honest, I have exactly the same concern as you. I thought about this the last few days, and I may have a potential workaround. How about adding a safety check at the beginning of the function: 
```julia
@assert strides(I) == (1, )
```

sshin23:
@frapac can you take a look?

src/IPM/callbacks.jl
```diff
@@ -1,9 +1,8 @@
-
 function eval_f_wrapper(ips::InteriorPointSolver, x::Vector{Float64})
     nlp = ips.nlp
     cnt = ips.cnt
     @trace(ips.logger,"Evaluating objective.")
-    cnt.eval_function_time += @elapsed obj_val = (get_minimize(nlp) ? 1. : -1.) * obj(nlp,view(x,1:get_nvar(nlp)))
+    cnt.eval_function_time += @elapsed obj_val = (get_minimize(nlp) ? 1. : -1.) * obj(nlp,unsafe_wrap(Vector{Float64},pointer(x),get_nvar(nlp)))
```
frapac, line 5: To improve readability, I would move the `unsafe_wrap` in a line before the function call. Something like
```julia
x_primal = unsafe_wrap(Vector{Float64},pointer(x),get_nvar(nlp))
```

src/KKT/sparse.jl
```diff
@@ -5,17 +5,17 @@
 Implement the [`AbstractReducedKKTSystem`](@ref) in sparse COO format.
 
 """
-struct SparseKKTSystem{T, MT} <: AbstractReducedKKTSystem{T, MT}
-    hess::StrideOneVector{T}
-    jac::StrideOneVector{T}
-    pr_diag::StrideOneVector{T}
-    du_diag::StrideOneVector{T}
+struct SparseKKTSystem{T, MT, VT} <: AbstractReducedKKTSystem{T, MT}
```
frapac, line 8: Good catch

src/KKT/sparse.jl
```diff
@@ -5,17 +5,17 @@
...
+    hess::VT
+    jac::VT
+    pr_diag::VT
+    du_diag::VT
     # Augmented system
-    aug_raw::SparseMatrixCOO{T,Int32}
+    aug_raw::SparseMatrixCOO{T,Int32,Vector{T}}
```
frapac, line 14: `Vector{T}` or `VT`?

frapac (commented):
It overall looks good to me! 
I think we might want to implement our own function on top of `unsafe_wrap`. For instance, something like: 
```julia
function _madview(vec::VT, n::Int, shift::Int=0)
    return unsafe_wrap(VT, pointer(vec) + shift, n)
end

```

PR #175, opened 2022-06-09
Define linear solver as a struct instead of as a module

frapac:
This PR defines the linear solver as proper Julia's struct, instead of a module. This allows to homogenize the code with the previous refactoring of the iterative refinement algorithm in #167 . 

### Summary 

In short, instead of defining a new MadNLP instance by passing a reference to the `MadNLPLapackCPU` module:
```julia
MadNLP.Optimizer(
    linear_solver=MadNLPLapackCPU,
)
```
this PR allows to instantiate MadNLP directly with the constructor of the linear solver:
```julia
MadNLP.Optimizer(
     linear_solver=LapackCPUSolver,
)
```

### To discuss

I am not sure about the naming of the different linear solvers (as previously they were all called `Solver`). Right now, I suggest:
- `MadNLPLapackCPU` -> `LapackCPUSolver`
- `MadNLPLapackGPU` -> `LapackGPUSolver`
- `MadNLPMumps` -> `MumpsSolver` 
-  `MadNLPMaXX` -> `MaXXSolver` 

But I welcome any feedback on this first suggestion!

### DIFF
--- a/lib/MadNLPKrylov/src/MadNLPKrylov.jl
+++ b/lib/MadNLPKrylov/src/MadNLPKrylov.jl
@@ -9,7 +9,8 @@ import IterativeSolvers:
     orthogonalize_and_normalize!, update_residual!, gmres_iterable!, GMRESIterable, converged,
     ModifiedGramSchmidt
 
-@kwdef mutable struct Options <: AbstractOptions
+
+@kwdef mutable struct KrylovOptions <: AbstractOptions
     krylov_restart::Int = 5
     krylov_max_iter::Int = 10
     krylov_tol::Float64 = 1e-10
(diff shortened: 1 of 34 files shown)


lib/MadNLPKrylov/src/MadNLPKrylov.jl
```diff
@@ -29,15 +30,15 @@ struct VirtualPreconditioner
 end
 ldiv!(Pl::VirtualPreconditioner,x::StrideOneVector{Float64}) = Pl.ldiv!(x)
 
-mutable struct Solver <: AbstractIterator
+mutable struct KrylovSolver <: AbstractIterator
```
sshin23, line 33: How about naming it as `KrylovIterator`?

src/options.jl
```diff
@@ -95,9 +94,9 @@ end
 end
 
 function check_option_sanity(options)
-    if options.linear_solver.INPUT_MATRIX_TYPE == :csc && options.kkt_system == DENSE_KKT_SYSTEM
-        error("[options] Sparse Linear solver is not supported in dense mode.\n"*
-              "Please use a dense linear solver or change `kkt_system` ")
-    end
+    # if options.linear_solver.INPUT_MATRIX_TYPE == :csc && options.kkt_system == DENSE_KKT_SYSTEM
```
sshin23, line 97: we should check it with `input_type()` function

sshin23 (approved):

PR #187, opened 2022-07-04
Add support for Float32

sshin23:
This PR adds support for `Float32`, or any other precision types.

To make this possible, we make the `AbstractLinearSovler` a parametric type, where the precision is given as a type parameter. The `Float32` version of the solver interfaces is added when it is available.


src/IPM/IPM.jl
```diff
@@ -176,22 +178,29 @@ function InteriorPointSolver{KKTSystem}(nlp::AbstractNLPModel, opt::Options;
 
     aug_vec_length = is_reduced(kkt) ? n+m : n+m+nlb+nub
 
-    _w1 = is_reduced(kkt) ? ReducedKKTVector(n, m) : UnreducedKKTVector(n, m, nlb, nub)
-    _w2 = is_reduced(kkt) ? ReducedKKTVector(n, m) : UnreducedKKTVector(n, m, nlb, nub)
-    _w3 = is_reduced(kkt) ? ReducedKKTVector(n, m) : UnreducedKKTVector(n, m, nlb, nub)
-    _w4 = is_reduced(kkt) ? ReducedKKTVector(n, m) : UnreducedKKTVector(n, m, nlb, nub)
+    if is_reduced(kkt)
+        _w1 =  ReducedKKTVector(similar(x,n+m), n, m)
```
frapac, line 182: Oups,  I think the syntax is a bit too complicated here. Should we create a new constructor for `ReducedKKTVector`? Something like
```julia
_w1 = ReducedKKTVector{T}(n, m)
```

src/IPM/IPM.jl
```diff
@@ -176,22 +178,29 @@ function InteriorPointSolver{KKTSystem}(nlp::AbstractNLPModel, opt::Options;
...
 
-    d = UnreducedKKTVector(n, m, nlb, nub)
+    d = UnreducedKKTVector(similar(x,n+m+nlb+nub), n, m, nlb, nub)
     dx_lr = view(d.xp, ind_cons.ind_lb) # TODO
     dx_ur = view(d.xp, ind_cons.ind_ub) # TODO
 
-    p = UnreducedKKTVector(n, m, nlb, nub)
+    p = UnreducedKKTVector(similar(x,n+m+nlb+nub), n, m, nlb, nub)
 
     obj_scale = [1.0]
```
frapac, line 201: Suggestion:
```julia
obj_scale = T[1.0] 
```

src/IPM/kernels.jl
```diff
@@ -311,10 +311,10 @@ function initialize_variables!(x,xl,xu,bound_push,bound_fac)
     end
 end
 
-function adjust_boundary!(x_lr,xl_r,x_ur,xu_r,mu)
+function adjust_boundary!(x_lr::A,xl_r,x_ur,xu_r,mu) where {T, A <: AbstractArray{T}}
```
frapac, line 314: Suggestion: rename the type `A` as `VT` to stick with the previous naming

src/LinearSolvers/linearsolvers.jl
```diff
@@ -37,6 +37,24 @@ factorized previously with [`factorize!`](@ref).
...
+
+# Examples
+```julia-repl
+julia> is_supported(UmfpackSolver,Float64)
+true
+
+julia> is_supported(UmfpackSolver,Float32)
+false
+```
+"""
```
frapac, line 53: Nice documentation!

src/LinearSolvers/linearsolvers.jl
```diff
@@ -37,6 +37,24 @@ factorized previously with [`factorize!`](@ref).
 """
 function solve! end
 
+"""
+    is_supported(solver,T)
+
+Return `true` if `solver` supports the Float type `T`.
```
frapac, line 43: Suggestion: 
```julia
Return `true` if `solver` supports the floating point number type `T`.
```

frapac (commented):
That's a significant improvement in MadNLP! I read this PR and had only a few minor comments so far. It would be interesting to find a good use-case to advertise this new capability. Maybe we should run the GPU benchmark on a simple GPU that does not support double precision?

sshin23:
Thanks, @frapac for the review! Indeed creating a good use case would be important, and probably it doesn't have a big advantage on CPU. 

Just ran a simple experiment on my laptop:
```julia
julia> T=Float32; N=400; a = CUDA.randn(T,N,N); a = a*a'+I; @time cholesky(a);
  0.000969 seconds (163 allocations: 9.016 KiB)

julia> T=Float64; N=400; a = CUDA.randn(T,N,N); a = a*a'+I; @time cholesky(a);
  0.002487 seconds (163 allocations: 9.016 KiB)
```
so, would be interesting to test the performance with a very large-scale dense problem on GPU. Would be interesting to test it with DynamicNLPModels.jl

cc: @dlcole3

lib/MadNLPPardiso/test/runtests.jl
```diff
@@ -20,6 +20,11 @@ testset = [
 ]
 
 @testset "MadNLPPardiso test" begin
+
+    MadNLPTests.test_linear_solver(PardisoMKLSolver,Float32)
+    MadNLPTests.test_linear_solver(PardisoMKLSolver,Float64)
+    # MadNLPTests.test_linear_solver(PardisoSolver)
```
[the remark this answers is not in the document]
sshin23: This can be uncommented once we get access to PardisoProject. Just wrote a comment about the current situation

PR #191, opened 2022-07-08
[Breaking change] Refactor the way we pass options to MadNLP

frapac:
This PR merges the `option_dict` and `kwargs` arguments, so now we pass all options as `kwargs...`:
```julia
ips = MadNLP.InteriorPointSolver(nlp; max_iter=10)
```
No need to pass the options in a separate `Dict` anymore. 

In addition, we add a new function `default_options` to return the default options associated to a given linear solver. 
```julia
MadNLP.default_options(LapackCPUSolver)
```

### DIFF
--- a/lib/MadNLPGPU/src/interface.jl
+++ b/lib/MadNLPGPU/src/interface.jl
@@ -1,41 +1,19 @@
 
-function CuInteriorPointSolver(nlp::AbstractNLPModel{T};
-    option_dict::Dict{Symbol,Any}=Dict{Symbol,Any}(), kwargs...
-) where T
-    # Initiate interior-point options
-    opt = MadNLP.IPMOptions(linear_solver=LapackGPUSolver)
-    linear_solver_options = set_options!(opt, kwargs)
-    MadNLP.check_option_sanity(opt)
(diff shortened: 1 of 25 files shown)


sshin23 (approved):
Thanks for the PR @frapac 

Since we introduce a breaking change, maybe it's an excellent time to change some function names. But we can do it in a separate PR.

lib/MadNLPGPU/src/interface.jl
```diff
@@ -2,10 +2,29 @@
...
     option_dict::Dict{Symbol,Any}=Dict{Symbol,Any}(), kwargs...
 ) where T
-    opt = MadNLP.Options(linear_solver=LapackGPUSolver)
-    MadNLP.set_options!(opt,option_dict,kwargs)
+    # Initiate interior-point options
+    opt = MadNLP.IPMOptions(linear_solver=LapackGPUSolver)
+    linear_solver_options = set_options!(opt, kwargs)
     MadNLP.check_option_sanity(opt)
 
+    # Initiate linear-solver options
```
sshin23, line 10: line 10-26 seems to be redundant to line 104-120 of IPM.jl. Can we create an internal function for setting options?

frapac:
@sshin23 thanks for the review! I am all in to change the names of the functions in a separate PR

PR #198, opened 2022-08-15
[hotfix] Symmetric `mul!` for dense/condensed KKT systems

sshin23:
Dense and condensed KKT systems perform incorrect matrix-vector multiplication when `aug_com` only stores the lower triangular entries. This PR fixes this issue


frapac (approved):
Good catch!

PR #200, opened 2022-08-15
add functions to decompose the time spent in callbacks/linear solver

frapac:


sshin23 (approved):
This looks handy! Thanks for the PR @frapac 

One question though, wouldn't it be better to return the result as a named tuple?

frapac:
right, named tuples should be more intuitive to use!

PR #208, opened 2022-08-30
Update quickstart.md

baggepinnen:
The name appears to have changed


frapac (approved):
Good catch, thanks for the fix!

baggepinnen:
There are actually several other things that appear to have changed
```
ips = MadNLP.MadNLPSolver(nlp)
ips.x
```
appears to contain both variables and Lagrange multipliers, and `ips.l` does not exist?

[Documenter.jl has a mechanism to run code in the docs and throw an error if it fails](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@example-block), it's a nice way to make sure the documentation is kept up to date.

PR #215, opened 2022-08-31
Fix `solve!` with `y` provided

### LINKED ISSUE #214: Possibly wrong init of dual variables
On the line
https://github.com/MadNLP/MadNLP.jl/blob/559bd21a12ca57b395076e244ea1334b69785c12/src/IPM/solver.jl#L122
the function `get_nvar` is called to get the length of the left-hand side, is this correct? I would expect the number of dual variables to be `get_ncon`?

In the same file
https://github.com/MadNLP/MadNLP.jl/blob/559bd21a12ca57b395076e244ea1334b69785c12/src/IPM/solver.jl#L11
the fields `meta.x, meta.y` are accessed, but they appear to be named
```
nlp.meta.x0, nlp.meta.y0
```

baggepinnen:
If this is indeed a correct fix, this closes #214

### DIFF
--- a/src/IPM/solver.jl
+++ b/src/IPM/solver.jl
@@ -119,7 +119,7 @@ function solve!(
         solver.x[1:get_nvar(solver.nlp)] .= x
     end
     if y != nothing
-        solver.y[1:get_nvar(solver.nlp)] .= y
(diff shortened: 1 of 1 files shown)


sshin23 (approved):
thanks for the PR @baggepinnen

PR #223, opened 2022-09-15
Replace `haskey` with `get!` to access buffers 

frapac:


lib/MadNLPGPU/src/kernels.jl
```diff
@@ -88,13 +88,9 @@ function MadNLP.jtprod!(y::AbstractVector, kkt::MadNLP.AbstractDenseKKTSystem{T,
...
     nx = size(kkt.jac, 2)
     ns = length(kkt.ind_ineq)
-    haskey(kkt.etc, :jac_w1) || (kkt.etc[:jac_w1] = CuVector{T}(undef, m))
-    haskey(kkt.etc, :jac_w2) || (kkt.etc[:jac_w2] = CuVector{T}(undef, nx))
-    haskey(kkt.etc, :jac_w3) || (kkt.etc[:jac_w3] = CuVector{T}(undef, ns))
-
-    d_x = kkt.etc[:jac_w1]::VT
-    d_yx = kkt.etc[:jac_w2]::VT
-    d_ys = kkt.etc[:jac_w3]::VT
+    d_x = get!(kkt.etc, :jac_w1, CuVector{T}(undef, m))
```
sshin23, line 91: I think this can cause unnecessary allocations. When `haskey(kkt.etc, :jac_w1) == true`, we don't want to create a new array. But it appears that this method allocates memory by creating `CuVector{T}(undef, m)` no matter `haskey(kkt.etc, :jac_w1)` is true or false.

sshin23:
Closing (see comment above)

Issue #186, opened 2022-07-02
Time-to-first-solve (TTFS)


frapac:
Now that the `StridedOneVector` PR #174 has been merged, we can focus on the time to first solve. 

Using the following script: 
```julia
using MadNLP
using MadNLPTests

nlp = MadNLPTests.HS15Model()
ips = MadNLP.InteriorPointSolver(nlp)
@time MadNLP.optimize!(ips)

``` 
MadNLP takes for the first solve: 
```
Total wall-clock secs in solver (w/o fun. eval./lin. alg.)  =  8.082
Total wall-clock secs in linear solver                      =  0.590
Total wall-clock secs in NLP function evaluations           =  0.037
Total wall-clock secs                                       =  8.709

EXIT: Optimal Solution Found.
  8.147811 seconds (24.28 M allocations: 1.187 GiB, 6.16% gc time, 99.93% compilation time)

```
It would be ideal if we can decrease that time to improve the user experience.

sshin23:
would be interesting to add benchmark code

sshin23:
@frapac Now that we support v1.9, might be good to revisit this and create a precompilation function

PR #227, opened 2022-09-30
[TTFS] Improve type inference in MadNLP

frapac:


frapac:
For information, this PR does not improve by a lot the precompilation time (apparently Julia is quite good to deal with this type of non-infered type). 

For the records, on MadNLP#master, we get with `SnoopCompile`: 
```
InferenceTimingNode: 5.120791/8.009915 on Core.Compiler.Timings.ROOT() with 9 direct children

```

On MadNLP#fp/type_inference: 
```
InferenceTimingNode: 4.226086/6.882473 on Core.Compiler.Timings.ROOT() with 2 direct children

```

x-ref: #186

sshin23 (approved):

PR #230, opened 2022-10-07
Remove allocations in callbacks 

frapac:
- add `PrimalVector` abstraction to remove the need to use `_madnlp_unsafe_wrapper` 
- rework KKT system to pass only the Jacobian w.r.t. the original variable `x` (without the slack) 
-

### DIFF
--- a/src/IPM/callbacks.jl
+++ b/src/IPM/callbacks.jl
@@ -119,7 +119,7 @@ function eval_lag_hess_wrapper!(solver::MadNLPSolver, kkt::AbstractDenseKKTSyste
     @trace(solver.logger,"Evaluating Lagrangian Hessian.")
     dual(solver._w1) .= l .* solver.con_scale
     hess = get_hessian(kkt)
-    scale = is_resto ? zero(T) : get_minimize(nlp) ? one(T) : -one(T)
+    scale = is_resto ? zero(T) : get_minimize(nlp) ? solver.obj_scale[] : -solver.obj_scale[]
     cnt.eval_function_time += @elapsed hess_dense!(
         nlp,
         variable(x),
--- a/src/IPM/callbacks.jl
+++ b/src/IPM/callbacks.jl
@@ -2,46 +2,48 @@ function eval_f_wrapper(solver::MadNLPSolver, x::Vector{T}) where T
     nlp = solver.nlp
     cnt = solver.cnt
     @trace(solver.logger,"Evaluating objective.")
-    x_nlpmodel = _madnlp_unsafe_wrap(x, get_nvar(nlp))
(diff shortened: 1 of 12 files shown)


src/IPM/callbacks.jl
```diff
@@ -1,128 +1,138 @@
...
     nlp = solver.nlp
     cnt = solver.cnt
     @trace(solver.logger,"Evaluating Lagrangian Hessian.")
-    dual(solver._w1) .= l.*solver.con_scale
+    dual(solver._w1) .= l .* solver.con_scale
     hess = get_hessian(kkt)
-    x_nlpmodel = _madnlp_unsafe_wrap(x, get_nvar(nlp))
-    hess_nlpmodel = _madnlp_unsafe_wrap(hess, get_nnzh(nlp.meta))
+    scale = (get_minimize(nlp) ? one(T) : -one(T))
+    scale *= (is_resto ? zero(T) : solver.obj_scale[])
```
sshin23, line 80: `scale = is_resto ? zero(T) : get_minimize(nlp) ? one(T) : -one(T)`

test/madnlp_test.jl
```diff
@@ -118,6 +118,30 @@ end
...
+    @test n_allocs == 16 # objective is still allocating
+    n_allocs = @allocated MadNLP.eval_grad_f_wrapper!(solver, f, x)
+    @test n_allocs == 0
+    n_allocs = @allocated MadNLP.eval_cons_wrapper!(solver, c, x)
+    @test n_allocs == 0
+    n_allocs = @allocated MadNLP.eval_jac_wrapper!(solver, kkt, x)
+    @test n_allocs == 0
+    n_allocs = @allocated MadNLP.eval_lag_hess_wrapper!(solver, kkt, x, solver.y)
+    @test n_allocs == 0
+end
```
sshin23, line 143: this is nice! maybe in the future we can do it for the whole solver level

src/KKT/sparse.jl
```diff
@@ -7,6 +7,7 @@ Implement the [`AbstractReducedKKTSystem`](@ref) in sparse COO format.
 """
 struct SparseKKTSystem{T, VT, MT} <: AbstractReducedKKTSystem{T, VT, MT}
     hess::VT
+    jac_callback::VT
```
sshin23, line 10: what does this do?

sshin23 (approved):
Nice work, @frapac! It overall looks great! I only have a few comments

test/madnlp_test.jl
frapac: I agree! we should at least apply the same test on the linear solver wrapper.

PR #232, opened 2022-10-17
Add new constructor for custom KKT type

frapac:
This new constructor allows to drastically simplify the support of custom KKT system outside MadNLP. 

Before, calling a custom KKT constructor was not straightforward, as illustrated by this code in Argos: 
https://github.com/exanauts/Argos.jl/blob/master/src/api.jl#L72-L80


sshin23 (approved):
Interesting. I'd love to see this feature used by others :)

PR #233, opened 2022-10-19
Improve MOI interface

sshin23:
This PR 
- adds support for `MOI.BarrierIterations`
- fixes the bug in `set_optimizer_attribute`

### DIFF
--- a/src/Interfaces/MOI_interface.jl
+++ b/src/Interfaces/MOI_interface.jl
@@ -172,7 +172,7 @@ end
 MOI.supports(::Optimizer, ::MOI.RawOptimizerAttribute) = true
 
 function MOI.set(model::Optimizer, p::MOI.RawOptimizerAttribute, value)
-    model.options[p.name] = value
+    model.options[Symbol(p.name)] = value
     # No need to reset model.solver because this gets handled in optimize!.
     return
 end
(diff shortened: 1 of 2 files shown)


src/Interfaces/MOI_interface.jl
```diff
@@ -689,6 +692,7 @@ function MOI.optimize!(model::Optimizer)
     model.solver = MadNLPSolver(model.nlp; model.options...)
     model.result = solve!(model.solver)
     model.solve_time = model.solver.cnt.total_time
+    model.solve_iterations = model.solver.cnt.k
```
frapac, line 695: How about using a getter instead? Something like `get_solve_iterations(model)`. 
The idea would be to avoid duplicating the information between `model.solve_iterations` and `model.solver.cnt.k`

src/Interfaces/MOI_interface.jl
```diff
@@ -883,3 +887,7 @@ function MOI.get(model::Optimizer, attr::MOI.NLPBlockDual)
     return s .* model.result.multipliers[(offset+1):end]
 end
 
+### MOI.BarrierIterations
+# This is impossible in Ipopt unless a custom callback function is used.
```
frapac, line 891: Maybe remove this comment as it does not apply here?

frapac (approved):
Looks good to me!

PR #235, opened 2022-10-24
Ma27 bug fix

sshin23:
This PR fixes the bug in ma27 code.

Because we can potentially resize `a`, which can potentially relocate `a`, we must not use `_madnlp_unsafe_wrapper`. This PR replaces it by `view(...)`.


frapac (approved):
Interesting catch. Out of curiosity, how have you noticed this issue?

sshin23:
@frapac was running a benchmark with ma27 :)

PR #236, opened 2022-11-09
Minor README link fix

michel2323:
Maybe also redirect https://github.com/sshin23/MadNLP.jl to here as Google still spits that out as the top link.


sshin23:
Thanks for reporting these @michel2323!

PR #241, opened 2023-01-29
Fix invalid return code in MOI.TerminationStatus

odow:
Bug reported on discourse: https://discourse.julialang.org/t/how-to-access-the-exit-of-madnlp-solver-directly/93685

### DIFF
--- a/src/Interfaces/MOI_interface.jl
+++ b/src/Interfaces/MOI_interface.jl
@@ -697,7 +697,7 @@ function MOI.optimize!(model::Optimizer)
 end
 
 # From Ipopt/src/Interfaces/IpReturnCodes_inc.h
-const _STATUS_CODES = Dict(
+const _STATUS_CODES = Dict{Status,MOI.TerminationStatusCode}(
     SOLVE_SUCCEEDED => MOI.LOCALLY_SOLVED,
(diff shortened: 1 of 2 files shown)


src/Interfaces/MOI_interface.jl
```diff
@@ -742,11 +753,7 @@ function MOI.get(model::Optimizer, ::MOI.TerminationStatus)
     elseif model.solver === nothing
         return MOI.OPTIMIZE_NOT_CALLED
     end
-    if haskey(_STATUS_CODES, model.result.status)
-        return _STATUS_CODES[model.result.status]
-    else
-        return MOI.UNKNOWN_RESULT_STATUS
```
odow, line 748: This was a `ResultStatusCode`, but it must be a `TerminationStatusCode`.

frapac (approved):
Thank you. The changes look good to me. The issue with the invalid values in the Hessian has been fixed in #243 . Once the tests are passing, I would suggest merging this PR before merging #243. We should update the test accordingly: 
https://github.com/MadNLP/MadNLP.jl/blob/fp/fix_nan_initial_multipliers/test/MOI_interface_test.jl#L80-L95

Issue #239, opened 2023-01-23
Invalid number in NLP Hessian Lagrangian detrected /  error in MadNLP


ChrisNabold:
using JuMP, MadNLP
m = Model(MadNLP.Optimizer)
@variable(m, x)
@variable(m, y)
@NLobjective(m, Min, (x-5)^2 + (y-8)^2)
@NLconstraint(m, x*y==5)
optimize!(m)
this small example errors in MadNLP. This example runs fine with Ipopt

frapac:
This is indeed an issue with MadNLP. 
When the initial point is not provided, the solver sets `x0` to 0. This is a problem in this case, as the Jacobian is null in 0, leading to the formulation of a degenerate KKT system. Then the KKT system is solved to estimate the initial multiplier (returning `NaN` in this degenerate case) which is itself use in the evaluation of the Hessian (inheriting the `NaN` values from the multiplier). As a result, MadNLP is returning an error. 

The quick fix is to provide an initial value to JuMP. In this case, MadNLP is returning the same solution as Ipopt:
```julia
using MadNLP
using JuMP

m = Model(Ipopt.Optimizer)
@variable(m, x, start=1.0)
@variable(m, y, start=1.0)
@NLobjective(m, Min, (x-5)^2 + (y-8)^2)
@NLconstraint(m, x*y==5)
optimize!(m)

```

But we have to address this issue properly in MadNLP. I see two workarounds: 
1. Use a dual regularization if the initial KKT system is degenerate to compute a valid initial multiplier. 
2. Assume directly the initial multiplier is equal to 0.0. 

I will investigate what Ipopt is doing in this case.

vinhpb:
Hi, 
I am using MadNLP with JuMP and also get this Invalid number in Hessian Lagrangian problem. When I tried to query the termination status with **termination_status(model)**, I got an error "MethodError: Cannot 'convert' an object of type MathOptInterface.ResultStatusCode to an object of type MathOptInterface.TerminationStatusCode". I could only see in the report that the error was due to this Invalid number in Hessian Lagrangian if I did not use **set_silent** to the JuMP model. Is there any way I can get this "Invalid number in Hessian Lagrangian" with something similar to termination_status(model)?
Thank you.

frapac:
Solved by #243

PR #243, opened 2023-01-31
Check whether initial KKT is valid before initiating multipliers

frapac:
Solve #239 
xref with #241


sshin23 (approved):
Looks good to me! Thanks for working on this @frapac 

Both #241 and #243 seem to be ready to merge

PR #259, opened 2023-05-11
[MadNLPTests] Remove ADNLPModels in deps

frapac:
ADNLPModels has too many dependencies (notably, Symbolics.jl) complicating the maintenance of MadNLPTests. 

Currently, we are using ADNLPModels only in the evaluation of the derivatives of the `NLSModel`. In this PR, we replace ADNLPModels by the analytical expressions of the derivatives.


sshin23 (approved):

sshin23:
Thanks @frapac for all the good old hand-written derivatives 😆

PR #261, opened 2023-05-24
Update for julia v1.9

sshin23:


sshin23:
The issue seems to be just cron not picking up the env variable. Now the test passes

PR #262, opened 2023-06-07
Using HSL_jll instead of custom compile

sshin23:
This PR updates MadNLPHSL.jl to use HSL_jll instead of the custom compiled library


sshin23:
not sure why PR is not updating -- will create a new one

PR #263, opened 2023-06-07
Using HSL_jll instead of custom compile

sshin23:
This PR updates MadNLPHSL.jl to use HSL_jll instead of the custom-compiled library


sshin23:
locally confirmed that HSL works on windows

frapac:
That looks good to me, it's good to see that HSL_jll is simplifying code a lot. 

Before merging this, I think it would be nice to run a benchmark, just to check the compilation flags are set properly in HSL_jll and we are not introducing any performance regression.

sshin23:
@frapac Thanks for the quick review!

Regarding the potential performance regression, I'd say we don't need to worry to much as the user can custom compile HSL_jll if they want to compile with specific compilation flags. Then, they can link it with Julia using `Overrides.jl` via the Dummy HSL_jll package https://juliahub.com/ui/Packages/HSL_jll/BQ0xV/1.0.0+0

cc @amontoison

lib/MadNLPHSL/src/ma86.jl
```diff
@@ -132,7 +132,11 @@ function Ma86Solver(
     csc::SparseMatrixCSC{T,Int32};
     opt=Ma86Options(),logger=MadNLPLogger(),
 ) where T
-    ma86_set_num_threads(opt.ma86_num_threads)
+
+    # Note: the current version of HSL_jll does not support openmp.
+    # this will be reenabled once openmp is supported in HSL_jll.
```
amontoison, line 137: `OpenMP` is not supported?!
Did I do a mistake when I compiled HSL_jll.jl?

amontoison:
For information, the `build_tarballs.jl` will be available tomorrow to regenerate an `HSL_jll.jl` from JuliaHSL.

lib/MadNLPHSL/src/ma86.jl
amontoison: @sshin23 I fixed the issue upstream.
I linked with `-fopenmp` but I didn't compiled with `-fopenmp`.

sshin23: Thanks for the quick action @amontoison! Indeed we used to have `-fopenmp` flag in our previous compilation script 
https://github.com/MadNLP/MadNLP.jl/blob/29e95a5e4f5806778524dc576b8f7223f076bbdf/lib/MadNLPHSL/deps/build.jl#L118
If you want to check it before releasing it, I'd be happy to test it with MadNLPHSL

amontoison:
@sshin23 
For the version of HSL_jll.jl precompiled with LBT, you need to load at least one BLAS / LAPACK backend.
We do that automatically in HSL.jl and Ipopt.jl:
https://github.com/JuliaSmoothOptimizers/HSL.jl/blob/main/src/HSL.jl#L16-L23

lib/MadNLPHSL/src/ma86.jl
sshin23:
```
x86_64-linux-gnu-libgfortran5-cxx11-libstdcxx30-julia_version+1.9.1
```

lib/MadNLPHSL/src/MadNLPHSL.jl
```diff
@@ -7,6 +7,16 @@ import MadNLP: @kwdef, MadNLPLogger, @debug, @warn, @error,
     get_tril_to_full, transfer!, input_type, _madnlp_unsafe_wrap,
     is_supported, default_options
 import HSL_jll: libhsl
+import LinearAlgebra, OpenBLAS32_jll
+
+function __init__()
+    if VERSION ≥ v"1.7"
```
amontoison, line 13: It's ```if VERSION ≥ v"1.9"```
I need to fix it in `HSL.jl`.

PR #266, opened 2023-06-12
bump MadNLP version to 0.7.0

sshin23:


frapac (approved):
It's the same issue as usual regarding the documentation: the build is failing because we need to use the latest version of MadNLPTests, not released yet.

I would suggest to discard this breaking test and merge this PR as is.

PR #277, opened 2023-11-13
[MadNLPHSL] Use HSL.jl

amontoison:
I started to remove the Julia wrappers but for some C structures (ma97_control, ma77_info, etc...) we should switch to the one defined in `HSL.jl`.

Note that with the dummy `HSL_jll.jl` available for everyone, `MadNLPHSL` could be merged in `MadNLP` because we can check if we can use the HSL solvers with `LIBHSL_isfunctional()`


sshin23:
Thanks @amontoison for working on this. I was wondering if it would be possible for HSL.jl to implement some initialization of `HSL.ma**_info` and `HSL.ma**_control` using keyword arguments so that `HSL.ma**_info` and `HSL.ma**_control` can be created without having to specify all the values. I think this can further specify interfacing with HSL solvers

amontoison:
Hi @sshin23,

Yes, we can do that. Can you open an issue or a small PR in HSL.jl?

amontoison:
@sshin23 
Do you know that you can easily create a a `ma**_control` with `new()` and after call the routine `default_control`?

sshin23:
An issue opened in HSL.jl [@amontoison Ah, you mean ](https://github.com/JuliaSmoothOptimizers/HSL.jl/issues/202)

sshin23:
@amontoison Ah, you mean [incomplete initialization](https://docs.julialang.org/en/v1/manual/constructors/#Incomplete-Initialization)? I think that should work as well

amontoison:
@sshin23 I did a new release of `HSL.jl` and updated `MadNLPHSL.jl`.

sshin23 (approved):
Thanks @amontoison for working on this!

PR #285, opened 2023-12-13
[API] Simplify API of SparseCallback and DenseCallback

frapac:
The idea is to simplify the manipulation of KKT systems outside MadNLP: previously we had to allocate a lot of 
structures to instantiate a KKT system using `create_kkt_system`: cf e.g. https://github.com/exanauts/Argos.jl/blob/fp/madnlp/test/Algorithms/MadNLP_wrapper.jl#L87-L111

* make explicit the options passed to `get_index_constraints`
* make explicit the options passed to `create_callback` and `initialize`
* add docstring for the structures in `src/nlpmodels.jl`


sshin23 (approved):
Looks great! Thanks @frapac

PR #290, opened 2024-01-03
[Algorithm] Improve LBFGS performance

frapac:
When running the benchmark for the MadNLP paper, I noticed the performance of MadNLP's LBFGS was behind Ipopt's LBFGS. Investigating it further, I noticed that: 
- LBFGS was not taking into account the fixed variables when present; 
- LBFGS was not scaling the constraints properly when calling `jtprod!`; 
- a check was missing to discard the update if the vectors `sk` or `yk` are null (leading to `NaN` in the algorithm and ugly failures). 
- LBFGS matrix was not initialized properly at iteration 1. 

This PR fixes all these issues. In addition, we made the following changes 
- reset the LBFGS matrix if too many updates are skipped (similar behavior as in Ipopt) ; 
- add new options to the algorithm (`sigma_min`, `sigma_max`, `init_value`) 
- add Gilbert & Lemaréchal rule to initialize the LBFGS matrix at the first iteration. 

With all these changes, we improve by a significant margin the robustness and the performance of the LBFGS algorithm. Comparing the performance with Ipopt's LBFGS on the CUTEst benchmark (instances with less than 1,000 variables), we get the following performance profile (in term of number of iterations):  

cutest-performance-profile-lbfgs

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ MadNLP is interfaced with modeling packages:
 - [JuMP](https://github.com/jump-dev/JuMP.jl)
 - [NLPModels](https://github.com/JuliaSmoothOptimizers/NLPModels.jl).
 
-Users can pass various options to MadNLP also through the modeling packages. The interface-specific syntax are shown below. To see the list of MadNLP solver options, check the [OPTIONS.md](https://github.com/MadNLP/MadNLP/blob/master/OPTIONS.md) file.
+Users can pass various options to MadNLP also through the modeling packages. The interface-specific syntax are shown below. To see the list of MadNLP solver options, check the [OPTIONS.md](https://github.com/MadNLP/MadNLP.jl/blob/master/OPTIONS.md) file.
 
 #### JuMP interface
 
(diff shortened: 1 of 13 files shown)


sshin23 (approved):
Thanks, @frapac. I wasn't able to check the details of the new initialization part, but the changes seem to be contained within LBFGS. So, as long as the performance is improved, I agree with the changes. Btw, can the benchmark script be added to the benchmark directory?

frapac:
I used the script in MadNLPBenchmarks.jl: https://github.com/MadNLP/MadNLPBenchmark.jl/blob/main/cutest/lbfgs/benchmark.jl

If you want, I could include it in the local `benchmark` directory as well. 

Long term, do you think we should keep the `benchmark` directory in MadNLP, or should we move all the benchmarks into MadNLPBenchmarks.jl?

sshin23:
@frapac indeed it would be better to separate the benchmark script in `MadNLPBenchmark.jl`. No need to include it here. Looks good to merge

PR #300, opened 2024-03-06
LDL factorization improvement

sshin23:
This PR improves LDLFactorization interface.

We also add a test for Float128 problems.

### DIFF
--- a/lib/MadNLPTests/src/Instances/hs15.jl
+++ b/lib/MadNLPTests/src/Instances/hs15.jl
@@ -1,9 +1,9 @@
-struct HS15Model <: NLPModels.AbstractNLPModel{Float64,Vector{Float64}}
-    meta::NLPModels.NLPModelMeta{Float64, Vector{Float64}}
+struct HS15Model{T} <: NLPModels.AbstractNLPModel{T,Vector{T}}
+    meta::NLPModels.NLPModelMeta{T, Vector{T}}
     counters::NLPModels.Counters
 end
 
-function HS15Model(; x0=zeros(2), y0=zeros(2))
+function HS15Model(T = Float64; x0=zeros(T,2), y0=zeros(T,2))
     return HS15Model(
         NLPModels.NLPModelMeta(
             2,     #nvar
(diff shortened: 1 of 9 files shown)


lib/MadNLPTests/src/Instances/hs15.jl
```diff
@@ -1,9 +1,9 @@
-struct HS15Model <: NLPModels.AbstractNLPModel{Float64,Vector{Float64}}
-    meta::NLPModels.NLPModelMeta{Float64, Vector{Float64}}
+struct HS15Model{T} <: NLPModels.AbstractNLPModel{T,Vector{T}}
+    meta::NLPModels.NLPModelMeta{T, Vector{T}}
     counters::NLPModels.Counters
 end
 
-function HS15Model(; x0=zeros(2), y0=zeros(2))
+function HS15Model(T = Float64; x0=zeros(T,2), y0=zeros(T,2))
```
frapac, line 6: I would maybe pass `T` directly in the optional arguments after the semicolon `;`?

src/LinearSolvers/cholmod.jl
```diff
@@ -5,7 +5,7 @@ end
 mutable struct CHOLMODSolver{T} <: AbstractLinearSolver{T}
     inner::CHOLMOD.Factor{Float64}
     tril::SparseMatrixCSC{T,Int32}
-    full::SparseMatrixCSC{Float64,Int32}
+    full::SparseMatrixCSC{Float64,Int}
```
frapac, line 8: Just ouf of curiosity, why do we have to use `Int` there?

frapac (approved):
I am impressed by how few changes are required to make `QuadMath` works within MadNLP. Good job! 

I just have one comment, but the PR looks good to merge.

src/LinearSolvers/cholmod.jl
sshin23: For `VERSION <= v"1.9"`, cholmod throws an error if the argument is `full::SparseMatrixCSC{Float64,Int32}
`

frapac: ah, good catch

PR #314, opened 2024-03-27
CuDSS synchronize added

sshin23:
Now we synchronize `CUDABackend` after factorize/solve with CUDSS. This fixes the timing inconsistency issue

### DIFF
--- a/lib/MadNLPGPU/src/cudss.jl
+++ b/lib/MadNLPGPU/src/cudss.jl
@@ -63,12 +63,16 @@ function MadNLP.factorize!(M::CUDSSSolver)
     CUDSS.cudss_set(M.inner.matrix, SparseArrays.nonzeros(M.tril))
     CUDSS.cudss("factorization", M.inner, M.x_gpu, M.b_gpu)
 
+    synchronize(CUDABackend())
+
     return M
 end
 
 function MadNLP.solve!(M::CUDSSSolver{T}, x) where T
     CUDSS.cudss("solve", M.inner, M.x_gpu, x)
     copyto!(x, M.x_gpu)
+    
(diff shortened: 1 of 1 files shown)


lib/MadNLPGPU/src/cudss.jl
```diff
@@ -63,12 +63,16 @@ function MadNLP.factorize!(M::CUDSSSolver)
...
+    synchronize(CUDABackend())
+
     return M
 end
 
 function MadNLP.solve!(M::CUDSSSolver{T}, x) where T
     CUDSS.cudss("solve", M.inner, M.x_gpu, x)
     copyto!(x, M.x_gpu)
+    
+    synchronize(CUDABackend())
```
frapac, line 75: I would assume the `copyto!` operation is already calling synchronize: 
https://github.com/JuliaGPU/CUDA.jl/blob/master/src/array.jl#L616-L617
even if some remarks shed some doubts on the consistency on more complicated workflows: 
https://github.com/JuliaGPU/CUDA.jl/blob/master/src/array.jl#L637-L638

frapac (approved):
LGTM! I just made a comment, but that's also something we can address in a separate PR.

lib/MadNLPGPU/src/cudss.jl
sshin23: Hmm indeed synchronize is already called. Is `CUDA.synchronize()` equivalent to `KA.synchronize(CUDABackend())`?

sshin23: Just confirmed that this synchronize is not necessary. We only need to synchronize after factorization

PR #317, opened 2024-04-05
Add an ordering for cuDSS

amontoison:
@frapac @sshin23 @michel2323

### DIFF
--- a/lib/MadNLPGPU/src/cudss.jl
+++ b/lib/MadNLPGPU/src/cudss.jl
@@ -2,8 +2,10 @@ import CUDSS
 import SparseArrays
 
 @kwdef mutable struct CudssSolverOptions <: MadNLP.AbstractOptions
-    # Use LDL by default in CUDSS as Cholesky can lead to undefined behavior. 
+    # Use LDLᵀ by default in CUDSS as Cholesky can lead to undefined behavior.
     cudss_algorithm::MadNLP.LinearFactorization = MadNLP.LDL
+    ordering::ORDERING = DEFAULT_ORDERING
+    perm::Vector{Cint} = Cint[]
 end
 
 mutable struct CUDSSSolver{T} <: MadNLP.AbstractLinearSolver{T}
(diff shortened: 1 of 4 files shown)


lib/MadNLPGPU/src/cudss.jl
```diff
@@ -20,6 +20,7 @@ function CUDSSSolver(
     csc::CUSPARSE.CuSparseMatrixCSC{T};
     opt=CudssSolverOptions(),
     logger=MadNLP.MadNLPLogger(),
+    ordering::Symbol=:default
```
sshin23, line 23: This should be moved to `CXudssSolverOptions` to be exposed to the users

sshin23: And maybe we can use https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/src/cusolverrf.jl#L184-L187
rather than symbols

sshin23 (commented):
Thanks, @amontoison! I only have small comments

lib/MadNLPGPU/src/cudss.jl
amontoison: The issue is that `CUCHOLESKYORDERING` has only two values. We can't add other ordering like `symamd` or `colamd` nor use the default ordering that is computed by cuDSS.

amontoison:
@sshin23 How can I test this new option with the following tests:
https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/test/madnlpgpu_test.jl#L4-L8
?

amontoison:
We have a segmentation fault with cuDSS if we provide our own permutation... :(

sshin23:
Hmm... maybe1-based indexing is causing an issue again? If it is possible to turn off the ordering, manual ordering is always an option.

https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/src/cusolverrf.jl#L237-L252

amontoison:
> Hmm... maybe1-based indexing is causing an issue again? If it is possible to turn off the ordering, manual ordering is always an option.
> 
> https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/src/cusolverrf.jl#L237-L252

cuDSS uses METIS internally but we want to compare the linear solvers on CPU / GPU with the same ordering.
The goal of this PR is to provide a permutation determined outside cuDSS.

sshin23:
Right. That can be implemented by turning off the ordering from the CUDSS side (if possible) and manually reordering the matrix before sending it to CUDSS, based on whatever ordering method we want to apply

amontoison:
> Right. That can be implemented by turning off the ordering from the CUDSS side (if possible) and manually reordering the matrix before sending it to CUDSS, based on whatever ordering method we want to apply

We can't turn off the ordering with the version 0.1.0 or 0.2.0. I will send an email to ask this feature for the next release.

amontoison:
@sshin23 @frapac 
The issue was that `SparseMatrixCSC` was not exported, so AMD/METIS were not called, and the permutation was `Cint[]`. CUDSS didn't like this permutation, and due to the segmentation fault, the error related to SparseMatrixCSC was not displayed...

lib/MadNLPGPU/test/madnlpgpu_test.jl
```diff
@@ -1,31 +1,49 @@
...
+        "CUDSS",
         ()->MadNLP.Optimizer(
             linear_solver=MadNLPGPU.CUDSSSolver,
             print_level=MadNLP.ERROR
         ),
         [],
     ],
     [
-        "LapackGPU-CUSOLVERRF",
+        "CUDSS",
```
sshin23, line 12: Maybe we can name them as `CUDSS`, `CUDSS-AMD`, `CUDSS-METIS`

sshin23 (approved):
Looks great! Thanks @amontoison for the contribution. I'll merge it once the naming is fixed

PR #321, opened 2024-04-10
[LinearSolvers] Add support for LDL factorization in CHOLMOD

frapac:
cc @amontoison

### DIFF
--- a/src/LinearSolvers/cholmod.jl
+++ b/src/LinearSolvers/cholmod.jl
@@ -23,6 +23,10 @@ function CHOLMODSolver(
     d = Vector{Float64}(undef,csc.n)
     full, tril_to_full_view = get_tril_to_full(Float64, csc)
 
+    if opt.cholmod_algorithm == LDL && VERSION <= v"1.9"
+        error("[CHOLMOD] Option `cholmod_algorithm=LDL` is not supported for Julia version <= 1.9")
+    end
+
     full = SparseMatrixCSC{Float64,Int}(
         full.m,
         full.n,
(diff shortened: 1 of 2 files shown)


sshin23 (approved):
LGTM! Have you observed any difference in performance?

amontoison (approved):

amontoison:
I think that they they did a major upgrade of SuiteSparse between Julia 1.10 (SuiteSparse 7.0.4) and Julia 1.9 (SuiteSparse 5.1.0).
Do we want to support both version of SuiteSparse or just bump the compat entry of Julia?

Off-topic: They added a function to provide the version number of CHOLMOD: https://github.com/JuliaSparse/SparseArrays.jl/commit/f890a1e13fc04c295cec48a9b08b76c6e1a7039a#diff-381fd98354d4dbfb825a400525529893f96061bad172b21f755066851fb1f49cR5

frapac:
> LGTM! Have you observed any difference in performance?

Currently running the benchmark on OPF instances ! Cholmod-LDL was slower than LDLFactorizations in my previous tests, but I was not using the correct inertia in MadNLP. 

> Do we want to support both version of SuiteSparse or just bump the compat entry of Julia?

Good point Alexis. I would suggest we support CHOLMOD-LDL only on Julia 1.10 then. What do you think would be best?

amontoison:
Yes, I think that it's best to simply add a check for `VERSION ≥ v"1.10"` and return an error that explicitly states that this linear solver is only supported with Julia 1.10.

sshin23:
Agreed with @amontoison's suggestion

sshin23 (approved):

amontoison (approved):

Issue #318, opened 2024-04-07
JuMP interface with `register`: `AssertionError: d.want_hess` error


franckgaga:
Related to #115

The `JuMP.jl` interface does not work with user-registered nonlinear operators. Here is a simple example running JuMP v1.21.0 and MadNLP v0.8.0:

```julia
using JuMP, MadNLP
optim = Model(MadNLP.Optimizer)
nvar = 2
function Jfunc(ΔŨtup...)
    ΔŨ = collect(ΔŨtup)
    return sum(ΔŨ.^2)
end
@variable(optim, ΔŨvar[1:nvar])
register(optim, :Jfunc, nvar, Jfunc, autodiff=true)
@NLobjective(optim, Min, Jfunc(ΔŨvar...))
optimize!(optim)
value.(ΔŨvar)
```
giving:
```
ERROR: AssertionError: d.want_hess
Stacktrace:
  [1] hessian_lagrangian_structure
    @ ~/.julia/packages/MathOptInterface/mz9FK/src/Nonlinear/ReverseAD/mathoptinterface_api.jl:296 [inlined]
  [2] hessian_lagrangian_structure(evaluator::MathOptInterface.Nonlinear.Evaluator{…})
    @ MathOptInterface.Nonlinear ~/.julia/packages/MathOptInterface/mz9FK/src/Nonlinear/evaluator.jl:179
  [3] hessian_lagrangian_structure(model::MadNLPMOI.Optimizer)
    @ MadNLPMOI ~/.julia/packages/MadNLP/waaOd/ext/MadNLPMOI/MadNLPMOI.jl:733
  [4] MadNLPMOI.MOIModel(model::MadNLPMOI.Optimizer)
    @ MadNLPMOI ~/.julia/packages/MadNLP/waaOd/ext/MadNLPMOI/MadNLPMOI.jl:839
  [5] optimize!(model::MadNLPMOI.Optimizer)
    @ MadNLPMOI ~/.julia/packages/MadNLP/waaOd/ext/MadNLPMOI/MadNLPMOI.jl:893
  [6] optimize!
    @ ~/.julia/packages/MathOptInterface/mz9FK/src/Bridges/bridge_optimizer.jl:380 [inlined]
  [7] optimize!
    @ ~/.julia/packages/MathOptInterface/mz9FK/src/MathOptInterface.jl:85 [inlined]
  [8] optimize!(m::MathOptInterface.Utilities.CachingOptimizer{…})
[...]

PR #322, opened 2024-04-10
[MOI] Add support for nonlinear problems without Hessian

frapac:
Solve https://github.com/MadNLP/MadNLP.jl/issues/318

This PR adds support for MOI nonlinear problems without Hessian. This is needed to support problems with user-defined functions (whose support has been improved in the latest MOI releases). Now MadNLP switches automatically to the LBFGS algorithm if the Hessian is not available, as it is done in Ipopt.jl. 

Currently, we specify to NLPModels that the problem does not have Hessian by setting `nnzh=0` in `NLPModelMeta`, so the following code works if the Hessian callback is not defined: 
https://github.com/MadNLP/MadNLP.jl/blob/fp/moi_nohessian/src/nlpmodels.jl#L360-L362
I am not sure this is the correct approach.

### DIFF
--- a/ext/MadNLPMOI/utils.jl
+++ b/ext/MadNLPMOI/utils.jl
@@ -201,24 +201,25 @@ function eval_sparse_gradient(
     f::MOI.ScalarQuadraticFunction{T},
     x::Vector{T},
     p::Dict{Int64,T},
+    adj::T,
 )::Int where {T}
     i = 0
     for term in f.affine_terms
         if !_is_parameter(term.variable)
             i += 1
-            ∇f[i] = term.coefficient
+            ∇f[i] += term.coefficient * adj
         end
     end
     for term in f.quadratic_terms
         if !_is_parameter(term.variable_1)
             v = _value(term.variable_2, x, p)
             i += 1
-            ∇f[i] = term.coefficient * v
+            ∇f[i] += term.coefficient * v * adj
         end
         if term.variable_1 != term.variable_2 && !_is_parameter(term.variable_2)
             v = _value(term.variable_1, x, p)
             i += 1
-            ∇f[i] = term.coefficient * v
+            ∇f[i] += term.coefficient * v * adj
         end
     end
     return i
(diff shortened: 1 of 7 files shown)


franckgaga:
Thanks for this PR! 
This a good compromise (clearly better than an error!) but according to JuMP stable documentation, both for the [NLP Legacy syntax](https://jump.dev/JuMP.jl/stable/manual/nlp/#Univariate-functions) and the (new NLP syntax)(https://jump.dev/JuMP.jl/stable/manual/nonlinear/#Gradients-and-Hessians), the Hessian is also automatically computed by AD with registered function/operators (with ForwardDiff.jl I think). LBFGS approximation should not required here.
Am I right @odow ?

odow:
> the Hessian is also automatically computed by AD with registered function

Only for univariate functions. We don't compute it for multivariate user-defined operators.

franckgaga:
Oops, my bad. My interpretation of the doc was incorrect. This is indeed the best solution in this case.

odow:
I think we kept the docs intentionally ambiguous, because it's a level of detail that shouldn't concern most people. The issue here is just a straight-up bug. You should check whether the AD backend supports `:Hess` before attempting to call it.

franckgaga:
@odow I'm curious, why JuMP chose to skip the Hessian for multivariate registered functions ? For the user using the legacy syntax, would you recommend to call `register` with a `∇²f`argument, and calling `ForwardDiff.hessian!` in the `∇²f` method?

ext/MadNLPMOI/utils.jl
```diff
@@ -201,24 +201,25 @@ function eval_sparse_gradient(
...
     x::Vector{T},
     p::Dict{Int64,T},
+    adj::T,
 )::Int where {T}
     i = 0
     for term in f.affine_terms
         if !_is_parameter(term.variable)
             i += 1
-            ∇f[i] = term.coefficient
+            ∇f[i] += term.coefficient * adj
```
sshin23, line 210: should this be `+=`? If we're evaluating non-compressed sparse grad, we wouldn't need `+=` here

sshin23 (approved):
LGTM. This reminds me of the possibility of piggybacking on Ipopt.jl/Knitro.jl...

odow:
> why JuMP chose to skip the Hessian for multivariate registered functions

Mainly because they can have many arguments, and computing the Hessian with ForwardDiff is expensive (because we'd need to compute the dense Hessian).

It would be breaking to enable this argument now.

The documentation has a tutorial on adding it manually if you're interested though:
https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/operator_ad/#Hessian-2

ext/MadNLPMOI/utils.jl
sshin23: NVM, I think it is possible that there is redundant entries

PR #325, opened 2024-04-12
Reexport MadNLP from MadNLP/libs

sshin23:
With this change (approach taken in [Makie](https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/src/CairoMakie.jl#L16-L22)), one can simply
```julia
using MadNLPGPU
```
instead of
```julia
using MadNLPGPU, MadNLP
```


frapac (approved):
LGTM!

PR #326, opened 2024-04-12
[MadNLPGPU] Bug fix for empty Hessian

sshin23:
This PR fixes the bug of throwing integer division error when dealing with empty Hessian

### DIFF
--- a/lib/MadNLPTests/src/MadNLPTests.jl
+++ b/lib/MadNLPTests/src/MadNLPTests.jl
@@ -111,7 +111,7 @@ end
 
 function test_madnlp(name,optimizer_constructor::Function,exclude; Arr = Array)
     @testset "$name" begin
-        for f in [infeasible,unbounded,lootsma,eigmina]
+        for f in [infeasible,unbounded,lootsma,eigmina,lp]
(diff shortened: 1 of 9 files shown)


lib/MadNLPTests/src/MadNLPTests.jl
```diff
@@ -331,6 +331,27 @@ function eigmina(optimizer_constructor::Function; Arr = Array)
     end
 end
 
+function lp(optimizer_constructor::Function; Arr = Array)
```
frapac, line 334: I would maybe add a comment referring to the github issue to explain why we have this test. How about renaming the function `lp_issueXXX` with `XXX` the id of the github issue?

frapac, line 334: I would maybe add a comment referring to the github issue to explain why we have this test. How about renaming the function `lp_issueXXX` with `XXX` the id of the github issue?

frapac (approved):
LGTM!

lib/MadNLPTests/src/MadNLPTests.jl
sshin23: Changed the name to `lp_examodels_issue75`

PR #339, opened 2024-05-14
Use GH Action Julia cache

michel2323:
This should speed up GH Actions CI and bump Julia to v1.10.


sshin23:
Interesting... Where are the caches stored?

michel2323:
Good question. In the cloud somewhere :-). It takes GH space on some balance. For moonshot I don't know.

Issue #349, opened 2024-06-12
[MadNLP] Improve adjustment of variable bound if slack is too small


frapac:
Distributed OPF leads to the formulation of degenerate nonlinear programs. Preliminary experiments on rosetta_distribution_opf show that MadNLP has different behavior than Ipopt. [This script](https://github.com/hei06j/rosetta_distribution_opf.jl/blob/main/script_4wire/IVR_EN_vectorized.jl) returns the following output with Ipopt: 
```shell
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  0.0000000e+00 2.25e-02 1.00e+00  -1.0 0.00e+00    -  0.00e+00 0.00e+00   0
   1  1.3084308e-02 5.95e-04 2.18e-01  -1.0 5.88e-02  -2.0 1.00e+00 9.80e-01h  1
   2  1.3411410e-02 8.64e-06 2.80e+00  -1.7 1.75e-03  -1.6 1.00e+00 9.86e-01h  1
   3  1.3412794e-02 1.19e-05 6.47e+03  -1.7 4.17e-05  -1.1 1.00e+00 2.99e-01f  1
   4  1.3416097e-02 7.18e-06 7.63e-03  -1.7 1.72e-05  -1.6 1.00e+00 1.00e+00f  1
   5  1.3416116e-02 2.86e-09 5.02e-05  -3.8 1.16e-05  -2.1 1.00e+00 1.00e+00h  1
In iteration 5, 1 Slack too small, adjusting variable bound
   6  1.3416116e-02 7.84e-09 3.35e+01  -8.6 2.57e-08  -2.6 1.00e+00 9.66e-01h  1
   7  1.3416116e-02 7.84e-09 1.21e+03  -8.6 2.03e-09  -3.1 1.00e+00 1.91e-06h 20
   8  1.3416116e-02 9.98e-09 2.41e-06  -8.6 2.22e-09  -3.5 1.00e+00 1.00e+00s 22
   9  1.3416116e-02 9.99e-09 2.02e-08  -8.6 1.83e-11  -4.0 1.00e+00 1.00e+00s 22

Number of Iterations....: 9
```
The slack is too small, so Ipopt adjusts the bound to avoid degenerate terms in the diagonal `S^{-1} V`. MadNLP does not do any adjustment on its side, and goes straight into numerical issue at iteration 5: 
```shell
[...]

PR #350, opened 2024-06-12
Check for NaN in linear solver's backsolve

frapac:
Throw an error in the iterative refinement algorithm if the residual has `NaN` . 
By throwing an error, we ensure that MadNLP exits the algorithm properly, and we avoid the behavior observed in #349 .


frapac:
Closing this PR, as it appears it introduces too many breaking changes (MadNLP is able to proceed to the next iteration even if some elements of the descent direction are `NaN`).

PR #356, opened 2024-07-16
[MadNLPGPU] Fix race conditions in GPU kernels

frapac:
Use `Atomix.@atomic` to avoid race conditions in custom GPU kernels.


sshin23:
huh, did we have race conditions before?

frapac:
Good question. 

I think race conditions might explain part of the non-deterministic runs we observed on the OPF and COPS benchmarks. But this  just an intuition, I don't have any elements to back this claim further. 

I noticed the issue with race-condition when porting MadNCL on the GPU. I observed a race condition right away and had to reimplement a few kernels.

sshin23:
Closing as this issue was mostly observed from MadNCL. A separate kernel may need to be developed on MadNCL side

Issue #229, opened 2022-10-06
Fixed variable multiplier


sshin23:
As noticed in #224, if a variable is fixed (its lower-bound equates its upper-bound) in MadNLP its multiplier is set to 0, which is wrong. The bound dual need to be properly set when terminating the solution.

Issue #70, opened 2021-09-04
Revisiting fixed variable treatment


sshin23:
Currently, fixed variable treatment is done by setting the associated RHS to zero and making the associated block in the augmented matrix identity. This should be changed to actually reducing the dimension of the KKT system.

PR #360, opened 2024-07-25
Revisit fixed variables treatment

frapac:
* Change the behavior of MakeParameter, depending on the callback:
   - `SparseCallback`: remove the fixed variables and reduce problem's dimension in MadNLP 
   - `DenseCallback`: keep the fixed variables frozen (as done previously). On the contrary to `SparseCallback`, we don't remove the fixed variables as this would require duplicating the Hessian and the Jacobian matrices to evaluate the model, which could be expensive if we are memory bounded. 
* Add a new `fixed_handler` if the problem has no fixed variable: `NoFixedVariable`.
* Ensure we always interact to the nonlinear program `nlp` through the callback wrapper `cb` to avoid side effect.
* Remove the function `get_index_constraint` and stores indexes directly in the callback wrapper `cb`.
* Fix the computation of the bound's multipliers for fixed variables.

Solve #70, #229

Warning: this PR adds a small breaking change in `create_kkt_system` (we remove `ind_cons` from the arguments) but I think it's worth the change, as the previous situation was error prone. Now, we make sure we store all the problem's information inside the callback wrapper `cb`. 

Preliminary results on `space_shuttle_reentry` shows that the current PR improves slightly the convergence (less iterations, less feasibility restoration), see the logs in attachment:
- [master.txt](https://github.com/user-attachments/files/16374217/space_param.txt)
- [this_pr.txt](https://github.com/user-attachments/files/16374225/space_param2.txt)

Benchmark on CUTEst is under way.

### DIFF
--- a/src/nlpmodels.jl
+++ b/src/nlpmodels.jl
@@ -1,3 +1,37 @@
+#=
+    MadNLP wrappers
+
+    MadNLP adapts any AbstractNLPModel to avoid numerical issues.
+    The `AbstractNLPModel` is wrapped as a `SparseCallback` or
+    as a `DenseCallback` (if the dense callbacks `jac_dense!` and `hess_dense!` are specified).
+
+    The wrapper reformulates the model by:
+    1. scaling the objective and the constraints.
+    2. removing the fixed variables from the formulation.
+
+    The scaling can be switched off by setting `nlp_scaling=false` in the options.
+
+    Four cases can occur for the fixed variables:
+
+    Case #1. The problem doesn't have any fixed variable
+        MadNLP sets `fixed_variable_treatment=NoFixedVariables()`
+        and fallbacks to the default callbacks (that just apply the scaling).
+
+    Case #2. `fixed_variable_treatment=RelaxBound`
+        MadNLP relax slightly the bounds of the fixed variables during
+        the initialization. The wrapper fallbacks to the default callbacks, similarly as with Case #1.
+
+    Case #3. `callback=DenseCallback` and `fixed_variable_treatment=MakeParameter`
+        Reformulate the fixed variables as dummy variables that are kept
+        at their bounds. The problem's dimension is not modified. The wrapper
+        modifies the Jacobian by filling the fixed columns with 0,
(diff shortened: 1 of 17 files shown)


frapac:
Benchmark is improved if we use the fix in #361 in this PR: 
[pprof-cpu.pdf](https://github.com/user-attachments/files/16412243/pprof-cpu.pdf)

src/nlpmodels.jl
```diff
@@ -286,33 +270,84 @@ function create_sparse_fixed_handler(
...
+    nx = length(free)
+    map_full_to_free = similar(jac_I, n) ; fill!(map_full_to_free, -1)
+    map_full_to_free[free] .= 1:nx
+
+    ind_jac_free = findall(@view(isfree[jac_J]))
+    ind_hess_free = findall(@view(isfree[hess_I]) .&& @view(isfree[hess_J]))
+
+    nnzh = length(ind_hess_free)
+    Hi, Hj = similar(hess_I, nnzh), similar(hess_J, nnzh)
+    Hi .= map_full_to_free[hess_I[ind_hess_free]]
```
sshin23, line 299: views and copyto here?

src/nlpmodels.jl
```diff
@@ -286,33 +270,84 @@ function create_sparse_fixed_handler(
...
     resize!(hess_J, nnzh)
-    resize!(hess_buffer, nnzh)
-    copyto!(@view(hess_I[end-nfixed+1:end]), fixed)
-    copyto!(@view(hess_J[end-nfixed+1:end]), fixed)
+    hess_I .= Hi
+    hess_J .= Hj
+
+    nnzj = length(ind_jac_free)
+    Ji, Jj = similar(jac_I, nnzj), similar(jac_J, nnzj)
+    Ji .= jac_I[ind_jac_free]
```
sshin23, line 308: views and copyto here?

sshin23 (approved):
LGTM. I just suggest we use copyto! Whenever we can. Thanks for working on this PR @frapac!

PR #361, opened 2024-07-25
Fix option `bound_relax_factor`

frapac:
- relax initial bounds with `bound_relax_factor` instead of `tol`. 
- significantly improve the performance of MadNLP on CUTEst 

[pprof-cpu.pdf](https://github.com/user-attachments/files/16383748/pprof-cpu.pdf)


sshin23:
@frapac Have we tested this with LiftedKKTSystem on GPU? This may make lifted KKT system more difficult to solve

PR #362, opened 2024-07-26
Make library relocatable

tmmsartor:
Me and @jgillis are developing a [C interface for MadNLP ](https://github.com/jgillis/madnlp_c/tree/relocatable), so that it can be bundled in a "standalone" shared library and used from non-Julia tools and frameworks. 
We are using PackageCompiler.jl to create the library, but we encountered some issues with relocatability, namely having the Julia depot path of the building machine hard-coded in some places.

Those commits seems to fix the problem in our build, even if I am not 100% sure this is the best way to solve the problem, given I am relatively new to Julia.

Would you be interested in making MadNLP.jl relocatable?


frapac:
Hi @tmmsartor , we found your project on github a week ago and are very curious about it. 

We would be in to make MadNLP relocatable, if that could make your life easier. Looking at your PR, it looks like you have identified two issues: 
1. we parse the toml file to get the current version of MadNLP
2. the MUMPS's interface is using `ccall` instead of `@call`
Is that right? 

Apparently RelocatableFolders.jl is needed only for 1. Let me check if we can find an alternative to using this new dependency.

src/MadNLP.jl
```diff
@@ -15,7 +16,9 @@ export MadNLPSolver, MadNLPOptions, UmfpackSolver, LDLSolver, CHOLMODSolver, Lap
 import LDLFactorizations
 
 # Version info
-version() = parsefile(joinpath(@__DIR__,"..","Project.toml"))["version"]
+# version() = parsefile(joinpath(@__DIR__,"..","Project.toml"))["version"]
+const PROJECT_TOML = @path joinpath(@__DIR__,"..","Project.toml")
```
frapac, line 20: Does the following work with PackageCompiler.jl? 
```julia
version() = packageversion(@__MODULE__)
```
Using that we don't have to parse the toml file and won't have to depend on RelocatableFolders.jl. 

This requires Julia >= 1.9

tmmsartor: I can try but I don't find this `packageversion` method is custom or part of some Package?

frapac: My apologies, I meant `pkgversion` instead of `packageversion`. The function is part of `Base`.

tmmsartor:
Yes I think those are the two issue we found.
I also think first issue can be solved without depending on RelocatableFolders.jl

Regarding the second issue I am not is the problem is `@ccall` vs `ccall` or the fact the artifacts lib path is appearing in an eval block. 
I notice that only for libmumps there was an hard-coded full path in the library, while other external libraries like cudss or metis were working fine, so I try to adapt the calling style to those cases.

frapac:
I think it's a good idea to remove the `@eval` block, and I am in to replace it by `@ccall`. 

If the tests pass and you confirm that you can use PackageCompiler.jl, we can merge this PR.

frapac (approved):
LGTM

tmmsartor:
I was not sure if there was a reason why in MadNLP is defined the lambda/method `version()` while in MadNLPMumps there is the variable `version`, to be safe I keep this distinction in the relocatability updates. 
Also  `pkgversion()` return `VersionNumber` while MadNLP tests seems to expect a string for `version()`/`version` so I added that conversion.
I was wondering if there is Julia standard or suggested way to provide this version information.

sshin23:
LGTM. merging it to master

PR #373, opened 2024-11-09
[CI] Setup buildkite

amontoison:
We could test `MadNLPGPU.jl` in a different PR.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,7 +3,7 @@ steps:
     plugins:
       - JuliaCI/julia#v1:
           version: "1.10"
-      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-test#v1: ~/lib/MadNLPGPU
     agents:
       queue: "juliagpu"
       cuda: "*"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,22 @@
+steps:
+  - label: "Julia v1.10"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.10"
+      - JuliaCI/julia-test#v1: ~
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
+    if: build.message !~ /\[skip tests\]/
+    timeout_in_minutes: 60
+
+  - label: "Julia v1.11"
+    plugins:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,22 @@
+steps:
+  - label: "Julia v1.10"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.10"
+      - JuliaCI/julia-test#v1: ~
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
```
sshin23, line 9: we don't need nvidia GPUs for MadNLPTesting (only needed for MadNLPGPU). Wouldn't this be waste of resources?

amontoison: It's a waste of ressources but I don't in which subpackages you are using `GPU`.

sshin23: We only need it for testing `MadNLPGPU`

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,22 @@
+steps:
+  - label: "Julia v1.10"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.10"
+      - JuliaCI/julia-test#v1: ~
```
amontoison, line 6:
```suggestion
      - JuliaCI/julia-test#v1: ~/lib/MadNLPGPU
```

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,22 @@
+steps:
+  - label: "Julia v1.10"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.10"
+      - JuliaCI/julia-test#v1: ~/lib/MadNLPGPU
```
amontoison, line 6:
```suggestion
      - JuliaCI/julia-test#v1: ~
```

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,22 @@
...
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
```
amontoison, line 9:
```suggestion
      cuda: "*"
    working_directory: "~/lib/MadNLPGPU"
```

sshin23:
Right now, I don't think we'll need buildkite as the current CI is working

PR #376, opened 2024-11-13
Upgrade MadNLPHSL.jl

amontoison:
Require the new release of `HSL_jll.jl` and this PR of `HSL.jl`:
https://github.com/JuliaSmoothOptimizers/HSL.jl/pull/215

@frapac You will be able to easily try it.

### DIFF
--- a/lib/MadNLPHSL/Project.toml
+++ b/lib/MadNLPHSL/Project.toml
@@ -1,6 +1,6 @@
 name = "MadNLPHSL"
 uuid = "7fb6135f-58fe-4112-84ca-653cf5be0c77"
-version = "0.5.2"
+version = "0.6.0"
 
 [deps]
 HSL = "34c5aeac-e683-54a6-a0e9-6e0fdc586c50"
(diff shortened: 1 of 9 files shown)


sshin23:
This is great! @amontoison, does only ma27 support quad precision? I guess it might be tricky once they rely on BLAS

amontoison:
Quadruple precision will only be supported in `MA27` for the upcoming release of `HSL_jll.jl`.
However, the next release will bring other useful features, such as Int64 support for the HSL solvers interface in `MadNLP`.

For supporting quadruple precision in other HSL linear solvers, we have a few challenges.
The first is the convention for BLAS/LAPACK symbols, which isn't standardized in quadruple precision:
- https://github.com/Reference-LAPACK/lapack/issues/1072

Next, I need to add these symbols to LBT, as some of them are missing:
- https://github.com/JuliaLinearAlgebra/libblastrampoline

Finally, we need a BLAS/LAPACK library compiled in quadruple precision with the relevant symbols.

@nimgould plans to port the subset of routines needed in quadruple precision:
- https://github.com/ralna/GALAHAD/issues/337

We might be able to include this in `HSL_jll.jl`, but all of this is speculative and will take some time before it becomes available.

sshin23:
Looks good @amontoison! Now I think it is a good time to get rid of `Int32` from MadNLP and make `Int64` the default (but in a separate PR)

sshin23 (approved):

amontoison:
MA27 can also be used with quadruple precision with PR.

I tested with the future version of HSL_jll.jl 2025.x.x that support Float128 for all solvers and it works too. I did the modifications in HSL.jl such that we don't need to update `MadNLPHSL.jl` again.

The function `HSL.is_supported` will automatically return `true` with Float128.

amontoison:
@frapac Can I merge this PR?

frapac (approved):
Looks good to me, thank you for the contribution @amontoison !

PR #393, opened 2025-01-07
Iterative refinement tolerance tuning

sshin23:
This PR implements tuning for the iterative refinement. Previously, the tolerance was fixed to `1e-8`, but this is impractical for single precisions. In this PR, we implement an adaptive tuning based on the overall tolerance of the solver. For double precision, we have the same behavior as before, but for single precision, this changes the behavior and makes things faster (less IR steps). 

It turns out that this PR makes the LiftedKKTSystem as it uses higher tolerance.

### DIFF
--- a/src/LinearSolvers/backsolve.jl
+++ b/src/LinearSolvers/backsolve.jl
@@ -1,7 +1,7 @@
 @kwdef mutable struct RichardsonOptions <: AbstractOptions
     richardson_max_iter::Int = 10
-    richardson_tol::Float64 = 1e-10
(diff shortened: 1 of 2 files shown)


sshin23:
cc: @Franc-Z

sshin23:
@Franc-Z I'm not sure if #382 is also related to this. Could you check if this resolves the issue in #382?

Franc-Z:
> @Franc-Z I'm not sure if #382 is also related to this. Could you check if this resolves the issue in #382?

No, it's not related to #382 , I have closed #382 yet.

frapac (approved):

PR #402, opened 2025-01-21
[MadNLPHSL] Add a self-hosted build

amontoison:
Test `MadNLPHSL` with the latest official release of `HSL_jll.jl` on a self-hosted runner (`moonshot`).
I compiled this `HSL_jll.jl` with debug options such we can easily track some errors.


amontoison:
It works :tada:

amontoison:
It seems that we have sometimes an overlap between the runners.
@sshin23 Do you still have access to Moonshot?
If not, I suggest to use my runner on Moonshot for all builds.

sshin23 (approved):

PR #403, opened 2025-01-23
[MadNLP] Fix the warning with SolverCore.solve!

amontoison:
Fix the following warning:
```julia
   1 dependency had output during precompilation:
┌ MadNLP
│  ┌ Warning: Replacing docs for `SolverCore.solve! :: Union{}` in module `MadNLP`
│  └ @ Base.Docs docs/Docs.jl:243
└  
```

### DIFF
--- a/src/KKT/KKTsystem.jl
+++ b/src/KKT/KKTsystem.jl
@@ -174,7 +174,7 @@ Solve the KKT system ``K x = w`` with the linear solver stored
 inside `kkt` and stores the result inplace inside the `AbstractKKTVector` `w`.
 
 """
-function solve! end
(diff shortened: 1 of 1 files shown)


sshin23:
Yes, this was annoying. Thank you

sshin23:
Would it make sense to `import SolverCore.solve!` and extend it?

amontoison:
> Would it make sense to `import SolverCore.solve!` and extend it?

It's what we already do.
The issue was that a new docstring to the generic function `solve!` was defined.

sshin23:
I see. Thank you @amontoison !

PR #406, opened 2025-02-02
CUDSS Cholesky fix and Cholesky new default

sshin23:
This PR uses the CUDSS's new Cholesky interface and makes Cholesky the default option


amontoison:
@sshin23 You will need the release 0.4.3 of CUDSS.jl: https://github.com/JuliaRegistries/General/pull/124190

The other alternative is to call the low level wrapper directly:
```julia
val = Ref{Cint}(0)
nbytes = sizeof(val)
CUDSS.cudssDataSet(solver.data.handle, solver.data, "info", val, nbytes)
```

frapac (approved):

PR #414, opened 2025-03-31
[DNMY][MadNLPMumps] use MUMPS.jl as backend

frapac:
MadNLP is breaking with MUMPS_jll 5.7.3. This PR attempts to fix this issue by using MUMPS.jl as a backend (this would reduce the maintenance burden long term). 

A question before merging this PR: do we necessarily need MPI to instantiate MUMPS? The previous version of MadNLPMumps was not depending on MPI, but MUMPS.jl requires an explicit call to `MPI.Init()` to ensure we can call MUMPS: https://github.com/JuliaSmoothOptimizers/MUMPS.jl/blob/main/src/interface.jl#L29

### DIFF
--- a/lib/MadNLPMumps/Project.toml
+++ b/lib/MadNLPMumps/Project.toml
@@ -1,15 +1,19 @@
 name = "MadNLPMumps"
 uuid = "3b83494e-c0a4-4895-918b-9157a7a085a1"
-version = "0.4.1"
+version = "0.5.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
+MUMPS = "55d2b088-9f4e-11e9-26c0-150b02ea6a46"
 MUMPS_seq_jll = "d7ed1dd3-d0ae-5e8e-bfb4-87a502085b8d"
 MadNLP = "2621e9c9-9eb4-46b1-8089-e8c72242dfb6"
 OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 
 [compat]
-MUMPS_seq_jll = "~5.3, ~500.600"
+MPI = "0.20.22"
+MUMPS = "1.5.0"
+MUMPS_seq_jll = "~500.700"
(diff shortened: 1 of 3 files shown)


lib/MadNLPMumps/test/runtests.jl
```diff
@@ -11,11 +11,12 @@ testset = [
 ]
 
 @testset "MadNLPMumps test" begin
+    MPI.Init() # TODO: check if we really need MPI
+    comm = MPI.COMM_WORLD
```
sshin23, line 15: is this line necessary?

lib/MadNLPMumps/Project.toml
```diff
@@ -1,15 +1,19 @@
 name = "MadNLPMumps"
 uuid = "3b83494e-c0a4-4895-918b-9157a7a085a1"
-version = "0.4.1"
+version = "0.5.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
```
sshin23, line 7: It seems that we only use MPI in the testing. Do we need MPI as dependency?

sshin23 (approved):

amontoison:
It depends if you want to use the sequential version of parralel version of MUMPS.

frapac:
So far, I think we only need the sequential version of MUMPS.

frapac:
I see that MUMPS.jl depends only on MUMPS_jll. Alexis, do you think it would be better if we interface directly MUMPS_seq.jll?

amontoison:
Using `MUMPS.jl` and `MUMPS_jll.jl` is not a good idea in that case. We interfaced the parallel version, which is the reason of development of MUMPS.

sshin23:
Why did MUMPS_jll 5.7.3 cause an issue for MadNLP in the first place? We have only been using MUMPS_seq_jll

frapac:
I see. In that case I think we can close this PR. I still have to understand why MadNLPMumps has memory issue with the latest release of MUMPS (5.7.3)

amontoison:
> I see that MUMPS.jl depends only on MUMPS_jll. Alexis, do you think it would be better if we interface directly MUMPS_seq.jll?

Yes, it will be easier. You don't need all the MPI stuffs.

Issue #418, opened 2025-04-08
[MadNLPGPU] LapackGPUSolver is broken with CUDA.jl 5.7


frapac:
When running the tests for MadNLPGPU with CUDA.jl v5.7, I observe the following error:

```
 CUDA error: an illegal memory access was encountered (code 700, ERROR_ILLEGAL_ADDRESS)     
```

The issue does not arise with CUDA.jl v5.6. Investigating it more closely, the problem arises within LapackGPUSolver. If I comment out [these lines](https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/test/madnlpgpu_test.jl#L48-L83), the tests are working fine again.

frapac:
The issue is with the QR factorization. Here is a MWE: 
```julia
using LinearAlgebra
using MadNLP
using MadNLPGPU
using CUDA

A = rand(10, 10) + 5I
A = A' + A
x = rand(10)

dA = CuArray(A)
dx = CuArray(x)

opt = MadNLP.LapackOptions(; lapack_algorithm=MadNLP.QR)
f = MadNLPGPU.LapackGPUSolver(dA; opt=opt)

MadNLP.factorize!(f)
MadNLP.solve!(f, x)

```

frapac:
Closed by #419

PR #419, opened 2025-04-08
Fix MadNLPGPU with CUDA.jl v5.7

frapac:
Solve #418


amontoison:
Is it still working with older versions of CUDA.jl?
Just to know if we need to bump the compat entry in Project.toml.

frapac:
Good question @amontoison . Let me do a check with CUDA.jl v5.6

frapac:
Alexis, I confirm it works with CUDA.jl v5.6.

Issue #423, opened 2025-04-09
Float32 in BFGS


michel2323:
Using `Float32` in BFGS leads to:
```julia
ERROR: LoadError: MethodError: no method matching MadNLP.CompactLBFGS(::MadNLP.BFGSInitStrategy, ::Vector{…}, ::Vector{…}, ::Vector{…}, ::Vector{…}, ::Vector{…}, ::Float64, ::Float64, ::Float64, ::Int64, ::Int64, ::Int64, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Matrix{…}, ::Vector{…}, ::Vector{…}, ::Vector{…})
The type `MadNLP.CompactLBFGS` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  MadNLP.CompactLBFGS(::MadNLP.BFGSInitStrategy, ::VT, ::VT, ::VT, ::VT, ::VT, ::T, ::T, ::T, ::Int64, ::Int64, ::Int64, ::MT, ::MT, ::MT, ::MT, ::MT, ::MT, ::MT, ::MT, ::MT, ::MT, ::MT, ::VT, ::VT, ::VT) where {T, VT<:AbstractVector{T}, MT<:AbstractMatrix{T}}
   @ MadNLP /disk/mschanen/julia_depot/dev/MadNLP/src/quasi_newton.jl:208

Stacktrace:
 [1] create_quasi_newton(::Type{…}, cb::MadNLP.SparseCallback{…}, n::Int64; options::MadNLP.QuasiNewtonOptions)
   @ MadNLP /disk/mschanen/julia_depot/dev/MadNLP/src/quasi_newton.jl:242
 [2] create_kkt_system(::Type{…}, cb::MadNLP.SparseCallback{…}, ind_cons::@NamedTuple{…}, linear_solver::Type{…}; opt_linear_solver::MadNLP.UmfpackOptions, hessian_approximation::Type, qn_options::MadNLP.QuasiNewtonOptions)
   @ MadNLP /disk/mschanen/julia_depot/dev/MadNLP/src/KKT/Sparse/augmented.jl:60
 [3] MadNLPSolver(nlp::TorchNLPModel{…}; kwargs::@Kwargs{…})
[...]

PR #424, opened 2025-04-09
Fix floating-point for options

michel2323:
This should fix https://github.com/MadNLP/MadNLP.jl/issues/423


sshin23 (approved):

amontoison:
I think we should update all options in the same time or none of them. We also have the HSLoptions, etc...
I will be more fan of a cast of the options and a major refactor later.

PR #427, opened 2025-05-08
Support MUMPS v5.8.0

amontoison:
Support MUMPS `v5.8.0`.

Julia 1.9 is required by the artifact `MUMPS_seq_jll`, because it is cross-compiled with LBT.
We should have done with the previous release but I didn't checked the compat entry of Julia.


amontoison:
All tests passed, I merge the PR and do a new release.

PR #446, opened 2025-07-03
[MadNLPGPU] Expose more options in cuDSS wrapper

frapac:


amontoison (approved):
@frapac You forgot the option to activate matching.
Because of the prefix `cudss_`, we can break some packages but I think the only package concerned is `MadNCL.jl`.

PR #453, opened 2025-07-13
Norm replaced with custom-norm for julia v1.12

sshin23:
Due to https://github.com/JuliaGPU/CUDA.jl/issues/2811, we temporarily replace `norm()` with `mynorm()`, where `map` is used instead


amontoison (changes requested):
`mynorm` should call directly `CUBLAS.nrm2` on GPU.
Please add a dispatch in `MadNLPGPU.jl`.

PR #458, opened 2025-08-11
Don't support CUDA 5.8.3

amontoison:
CUDA 5.8.3 supports the toolkit v13.0 and they removed all the legacy API of CUSOLVER.
See https://github.com/JuliaGPU/CUDA.jl/pull/2842

### DIFF
--- a/lib/MadNLPGPU/Project.toml
+++ b/lib/MadNLPGPU/Project.toml
@@ -14,7 +14,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
 AMD = "0.5"
-CUDA = "5.4.0"
+CUDA = "5.4.0 - 5.8.2"
 CUDSS = "0.5.0"
 KernelAbstractions = "0.9"
 MadNLP = "0.8.8"


lib/MadNLPGPU/Project.toml
```diff
@@ -14,7 +14,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
 AMD = "0.5"
-CUDA = "5.4.0"
+CUDA = "5.4.0 - 5.8.2"
```
giordano, line 17: This is useless done here only, the only result will be that older versions of this package will be installed together with newer CUDA.jl.  You want to fix compat bounds in the registry as well.

sshin23:
Is this related? @amontoison 
```
julia> CUDA.zeros(Float64,3,30)
ERROR: Failed to compile PTX code (ptxas exited with code 255)
Invocation arguments: --generate-line-info --verbose --gpu-name sm_70 --output-file /tmp/jl_DYbS9CLqbe.cubin /tmp/jl_ryFirI709M.ptx
ptxas fatal   : Value 'sm_70' is not defined for option 'gpu-name'
If you think this is a bug, please file an issue and attach /tmp/jl_ryFirI709M.ptx
```

sshin23:
It looks like it does

```
julia> CUDA.zeros(Float64,10,10)
10×10 CuArray{Float64, 2, CUDA.DeviceMemory}:
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

julia> 

julia> 

(opf) pkg> st
...
⌃ [052768ef] CUDA v5.8.2
...
```

amontoison:
@sshin23 We need to open a PR in the general registry.
I did it for CUDSS.jl: https://github.com/JuliaRegistries/General/pull/136538/files

PR #470, opened 2025-08-18
primal residual comes from dual part of rhs and vice versa

apozharski:
Fix bug in adaptive barrier that flipped primal and dual residual.


frapac:
This PR has been merged manually.

PR #476, opened 2025-09-07
Add an extension for oneAPI.jl

amontoison:


sshin23:
@amontoison I'm starting to realize that there is quite a lot of redundant code between LapackCPU vs cuSOLVER vs rocSOLVER, etc. Wouldn't there be a better way?

amontoison:
I think we can rely more on multiple dispatch. I can propose a unified API next week.

The extension of oneAPI.jl is more a proof of concept than anything else because many things are still unstable in `oneAPI.jl`, but @michel2323 is officially working on it now!

michel2323:
Should we add an Intel GPU runner?

michel2323:
```ERROR: LoadError: UndefVarError: `oneSparseMatrixCSC` not defined in `oneAPI.oneMKL``` . Intel not like CSC.

sshin23:
@michel2323, we don't have an Intel runner. Could you add one?

michel2323:
Added oneAPI runner. @amontoison I forgot again how to resolve the CSR/CSC issue.

amontoison:
@michel2323 Add a CSC format in `oneAPI.jl` and use the relation between CSC / CSR to interface operations like `mul!`.

michel2323:
@amontoison I'm hitting now a `findall(::oneArray` in MadNLP, which is not supported in `oneAPI.jl`. While trying to implement that I've hit some indexing issue. I'm confident though that can be resolved.

michel2323:
@sshin23 After this, I can take care of making CUDA a MadNLPGPU extension too and try to consolidate the code for all three backends.

amontoison:
@michel2323 
A file `oneapi_qn.jl` is missing in the extension.
Do you have an idea why the tests are failing?

michel2323:
Because I forgot to commit the file 💩.

PR #481, opened 2025-09-15
Dual regularize diagonal

sshin23:
For `regular!`, `kkt.du_diag .= .-dual` is fine, as the dual reg is initially zero, but for restoration, this will reset the dual regularization, which involves the barrier term for `pp` and `nn` variables.


sshin23:
This is WIP. Still need to make sure it doesn't cause any unintended performance degradation. Since dual regularization is not strengthened each time, this should be ok, I think, but we need to run a benchmark.

sshin23:
EVD failure doesn't seem to be related

PR #482, opened 2025-09-17
Mumps tuned and made as default linear solver

sshin23:


src/LinearSolvers/mumps.jl
```diff
@@ -219,19 +197,14 @@ function MumpsSolver(csc::SparseMatrixCSC{T,Int32};
     mumps_struc.icntl = setindex(mumps_struc.icntl,opt.mumps_pivot_order,7)
     mumps_struc.icntl = setindex(mumps_struc.icntl,opt.mumps_scaling,8)
     mumps_struc.icntl = setindex(mumps_struc.icntl,0,10)
-    mumps_struc.icntl = setindex(mumps_struc.icntl,1,13)
+    mumps_struc.icntl = setindex(mumps_struc.icntl,0,13)
```
frapac, line 200: For reference, here are the options used in Ipopt: 
```c++
   mumps_data->icntl[5] = mumps_permuting_scaling_;
   mumps_data->icntl[6] = mumps_pivot_order_;
   mumps_data->icntl[7] = mumps_scaling_;
   mumps_data->icntl[9] = 0;   //no iterative refinement iterations

   mumps_data->icntl[12] = 1;   //avoid lapack bug, ensures proper inertia; mentioned to be very expensive in mumps manual
   mumps_data->icntl[13] = mem_percent_; //% memory to allocate over expected
   mumps_data->cntl[0] = pivtol_;  // Set pivot tolerance
```
Are we deactivating the IR in Mumps?

sshin23:
@frapac Yes, can confirm that it runs on single thread

KSepetanc:
If I understand correctly, this removes MadNLPMumps extension. Can you update docs? There is now broken link to that extension.

PR #488, opened 2025-09-23
[CUDSS] Return dummy inertia if factorization incomplete

sshin23:


amontoison:
@sshin23 With the last release of `CUDSS.jl`, we can easily recover the diagonal of the factorized matrix in-place.
I updated the code to explain how to do that, but kept it commented depending on how you want to use it.

sshin23:
@amontoison I was able to check the inertia calculation based on diagonal values, but it didn't seem to be too different from the one reported by CUDSS. So, I'll keep the current version. Thanks for the help

PR #490, opened 2025-09-25
option treatment and specialization for cudss

### LINKED ISSUE #438: [MadNLPGPU] Use CHOLESKY insted of LDL by default?
It is a breaking change so we should be careful if we do this modification.

sshin23:
close #438

This PR implements `default_options` for the linear solver that specializes in NLPModel and KKTSystem data types. With this feature, we specialize in the CUDSS option --- use Cholesky for LiftedKKT

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,6 +97,7 @@ jobs:
   # need to run one on cuda and one on amdgpu
   test-gpu:
     strategy:
+      fail-fast: false
       matrix:
         julia-version: ['1']
         gpu: [cuda, amdgpu]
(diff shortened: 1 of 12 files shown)


sshin23:
@frapac @amontoison it would be good to check if this will cause any unexpected behavior from MadNCL or MadIPM side. For LiftedKKT, Cholesky is ideal algorithm, so in this PR we implement a function `MadNLP.default_options(::AbstractNLPModel{T,VT}, ::Type{MadNLP.SparseCondensedKKTSystem}, linear_solver::Type{CUDSSSolver})`, which allows to specialize the default linear solver option depending on the choice of KKTSystem and NLPModel data type

amontoison:
It is a good idea! I wanted to find a nice way to use the option `"schur_mode"` by default for a KKT formulation based on the Schur complement in `MadIPM.jl`.
It seems that you provide it.

I will test it tomorrow or Monday.

sshin23:
We can also add solver to the argument if finer grained control is needed

PR #497, opened 2025-10-08
Overhaul solver member access via functions

apozharski:
The goal of this PR is to attempt to minimize the number of structure dependent `solver.*` calls in core MadNLP. 
The idea here is it easier in principle to create alternate subclasses of `AbstractMadNLPSolver` that only need to specialize some or all of the API found in `IPM/api.jl` and in doing so can make maximum use of the existing code in MadNLP with minimal required duplication.

I suspect there is more to be done in this direction, e.g. wrapping more "calculated" quantities over the current solver state, into functions `_quantity(solver::AbstractMadNLPSolver) = <expression>`, but I believe this is a good first pass.

(Note, that I am not married to the naming scheme etc., of this PR, if you would prefer `set_<field>(solver::AbstractMadNLPSolver)` and `get_<field>(solver::AbstractMadNLPSolver)`, I would be happy with that as well, however due to shadowing pain I think just `<field>(solver::AbstractMadNLPSolver)` is not the best idea.)

### DIFF
--- a/src/IPM/api.jl
+++ b/src/IPM/api.jl
@@ -26,6 +26,7 @@ set_obj_val_trial!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.o
 set_inf_pr!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.inf_pr = rhs
 set_inf_du!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.inf_du = rhs
 set_inf_compl!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.inf_compl = rhs
+set_inf_compl_mu!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.inf_compl_mu = rhs
 set_theta_min!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.theta_min = rhs
 set_theta_max!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.theta_max = rhs
 set_tau!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.tau = rhs
@@ -42,4 +43,5 @@ get_theta(solver::AbstractMadNLPSolver) = get_theta(get_c(solver))
 get_varphi(solver::AbstractMadNLPSolver) = get_varphi(get_obj_val(solver), get_x_lr(solver), get_xl_r(solver), get_xu_r(solver), get_x_ur(solver), get_mu(solver))
 get_kkt_error(solver::AbstractMadNLPSolver) = max(get_inf_pr(solver), get_inf_du(solver), get_inf_compl(solver))
 get_inf_compl(solver::AbstractMadNLPSolver{T}, sc::T; mu=get_mu(solver)) where {T} = get_inf_compl(get_x_lr(solver),get_xl_r(solver),get_zl_r(solver),get_xu_r(solver),get_x_ur(solver),get_zu_r(solver), mu, sc)
+get_inf_barrier(solver::AbstractMadNLPSolver) = max(get_inf_pr(solver),get_inf_du(solver),get_inf_compl_mu(solver))
 get_inf_total(solver::AbstractMadNLPSolver) = max(get_inf_pr(solver),get_inf_du(solver),get_inf_compl(solver))
--- a/src/IPM/api.jl
+++ b/src/IPM/api.jl
@@ -5,7 +5,7 @@
 
 # getters for all fields
 for (k, attribute) in enumerate(fieldnames(MadNLPSolver))
-    fname = "_$(attribute)"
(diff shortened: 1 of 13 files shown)


apozharski:
Just a note: I cannot reproduce the `test-madnlp` failure locally on my machine, which is odd.

sshin23:
@aspozharski could be due to the new release of julia v1.12

frapac (commented):
I like the direction where this is going. Having getters would also help with the C interface. 

I would suggest generating all the getters automatically using metaprogramming. E.g. using the following code:
```julia
for (k, attribute) in enumerate(fieldnames(MadNLPSolver))
    fname = "get_$(attribute)"
    @eval begin
        function $(Symbol(fname))(solver::MadNLPSolver)
            return getfield(solver, $k)
        end
    end
end

```

apozharski:
While I agree that the metaprogramming is nice from a maintenance perspective I always have found it to be a pain for future developers because things are not explicitly e.g. greppable.

apozharski:
> @aspozharski could be due to the new release of julia v1.12

Yep indeed there were some issues from the v1.12 release. #498 should fix these.

src/IPM/api.jl
```diff
@@ -0,0 +1,45 @@
+# API for getting internal members of IPM
+# The idea here is that a subtype of AbstractMadNLPSolver may specialize
+# some or all of these and the core functionality of the IPM in MadNLP
+# should function as normal.
+
+# getters for all fields
+for (k, attribute) in enumerate(fieldnames(MadNLPSolver))
+    fname = "_$(attribute)"
```
frapac: I would suggest prefixing the getters with `get_`: 
```julia
fname = "get_$(attribute)"
```

src/IPM/api.jl
```diff
@@ -0,0 +1,45 @@
...
+set_alpha!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.alpha = rhs
+set_alpha_z!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.alpha_z = rhs
+set_ftype!(solver::AbstractMadNLPSolver, rhs::String) = solver.ftype = rhs
+set_del_w!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.del_w = rhs
+set_del_w_last!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.del_w_last = rhs
+set_del_c!(solver::AbstractMadNLPSolver{T}, rhs::T) where {T} = solver.del_c = rhs
+
+# Computed quantities
+# TODO(@anton) I now think these would be useful to have, but which quantities are wrapped this way is a good question
+_theta(solver::AbstractMadNLPSolver) = get_theta(_c(solver))
```
frapac: I would be in to rely on multiple dispatch to define these new getters. Something like:
```julia
get_theta(solver::AbstractMadNLPSolver) = get_theta(get_c(solver))
```

frapac (changes requested):

src/IPM/api.jl
apozharski: Done!

src/IPM/api.jl
```diff
@@ -0,0 +1,45 @@
...
+    fname = "_$(attribute)"
+    @eval begin
+        @inline function $(Symbol(fname))(solver::MadNLPSolver)
+            return getfield(solver, $k)
+        end
+    end
+end
+
+# Necessary Mutators
+set_RR!(solver::AbstractMadNLPSolver, rhs) = solver.RR = rhs
```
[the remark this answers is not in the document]
frapac: that works for me!

frapac (approved):

src/IPM/kernels.jl
```diff
@@ -326,8 +326,8 @@ end
 
 function get_min_complementarity(solver::AbstractMadNLPSolver)
     return get_min_complementarity(
-        solver.x_lr, solver.xl_r, solver.zl_r,
-        solver.x_ur, solver.xu_r, solver.zu_r,
+        get_x_lr(solver), get_xl_r(solver), get_zl_r(solver),
+        get_x_ur(solver), get_xu_r(solver), get_zu_r(solver),
```
klamike, line 330: My comment during the call would essentially do
```suggestion
        lower(solver.x), get_xl_r(solver), get_zl_r(solver),
        upper(solver.x), get_xu_r(solver), get_zu_r(solver),
```

I think it's a non-issue/different scope. Better to push it off to a change in MadCore (then MadNLP can do `get_x_lr(solver::MadNLPSolver) = MadCore.lower(solver.x)` for example), so this PR stays single-focus

apozharski: Hmmm yea I think this is somewhat different scope now that I look at it. I think it is a good idea but I think having a followup for `PrimalVector` and co makes sense to me.

apozharski:
@sshin23 I think with the discussion above w.r.t the proposed changes during the meeting being pushed off to a separate PR, this is ready to go in with your approval.

PR #498, opened 2025-10-09
Fix tests broken in v1.12

apozharski:
This fixes tests broken by the Julia v1.12 release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,6 +97,7 @@ jobs:
   # need to run one on cuda and one on amdgpu
   test-gpu:
     strategy:
+      fail-fast: false
       matrix:
         julia-version: ['1']
         gpu: [cuda, amdgpu]


apozharski:
The remaining GPU failure seems to be related to https://github.com/JuliaGPU/AMDGPU.jl/issues/607 but this suggests that 1-norm should succeed. However, it seems in v1.12 this no-longer is true. The fix I implemented is simply ported from the corresponding fix for CuArray.

apozharski:
The remaining issues now seem to be caused by https://github.com/JuliaGPU/AMDGPU.jl/issues/827. This is claimed to be fixed then on master but it seems not on the latest release.

frapac:
I am merging this PR, even if the tests are not passing for amdgpu. 

The issue looks unrelated, and non-trivial to fix (see #499 ). I think it's more urgent to fix MadNLP on julia 1.12, now that the new version has been released.

PR #499, opened 2025-10-09
Fixes for AMDGPU

apozharski:
This is a PR which contains fixes for v1.12 for the AMDGPU extension. Separated out from #498.

### DIFF
--- a/lib/MadNLPGPU/Project.toml
+++ b/lib/MadNLPGPU/Project.toml
@@ -1,6 +1,6 @@
 name = "MadNLPGPU"
 uuid = "d72a61cc-809d-412f-99be-fd81f4b8a598"
-version = "0.7.14"
+version = "0.7.15"
 
 [deps]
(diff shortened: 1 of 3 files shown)


apozharski:
I think the core problem here is the versioning restriction on `CUDSS.jl` combined with incompatible versioning restrictions on `PrettyTables.jl`in `CUDA.jl < 5.9` and `AMDGPU = 2.1.1`.

Which may I say I _do not_ understand why two GPU libraries each have hard dependencies on a pretty printing library :exploding_head:

amontoison:
Please don't relax the compat bound on CUDA. It is to protect the user from installation of CuDA toolkit 13 where no installation of CUDSS is available (yet).

amontoison:
Ok we need a new release of `CUDSS.jl` such that we can upgrade `CUDA.jl` (> 5.8.2) first.

amontoison:
@apozharski New releases of `CUDSS.jl` and `MadNLPGPU.jl` are available.

PR #504, opened 2025-10-13
[WIP] extract code in `regular!`. `robust!`, and `restore!` into functions for more consise extensibility.

apozharski:
WIP until #497 goes in.

This PR separates out the steps within each MadNLP iteration as `step_name(solver::AbstractMadNLPSolver)`. Extensions to MadNLP then only need to provide specializations for the sections of the algorithm that they change. This should help minimize the maintenance burden for those extensions.

If people want to review just the changes in this PR easily the diff can be found here: https://github.com/apozharski/MadNLP.jl/pull/3/files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ benchmark/Project.toml
 benchmark/Manifest.toml
 benchmark/*.csv
 benchmark/*.pdf
+
+# Emacs Autosaves
+*~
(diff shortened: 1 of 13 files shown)


apozharski:
Closing because I am not 100% sold on this and out of band conversation with @frapac, w.r.t `MadCore` work.

I will probably return to this either after or during the `MadCore` implementation.

PR #509, opened 2025-10-23
Update the interface for MathOptInterface.jl -- support VectorNonlinearOracle

### LINKED ISSUE #508: [MOI] Support VectorNonlinearOracle
x-ref https://github.com/jump-dev/Ipopt.jl/pull/466
x-ref https://github.com/jump-dev/Ipopt.jl/pull/506

I will backport what I am doing in `UnoSolver.jl` here.
Last time, it was quite useful: https://github.com/MadNLP/MadNLP.jl/pull/493, https://github.com/MadNLP/MadNLP.jl/pull/494

amontoison:
close #508 

I updated a little the order of some routines such that the code is almost the same here and in `UnoSolver.jl`.
It will help to maintain the extension if bugs are found in one of the two Julia packages.

@frapac One the potential issue with `VectorNonlinearOracle` is that we don't have support for `jprod!`, `jtprod!` or `hprod!`. Do we always use `jtprod!` is `MadNLP.jl`? Do we have a workaround?


frapac:
superseded by #548

Issue #505, opened 2025-10-14
[Bug] Linear solver backend


ocots:
@frapac Hi! It seems to me that the following is a bug. In the following, the first solve is made with `umfpack` linear solver while the second with `mumps`. Actually, I was expecting `mumps` for both. To get `mumps` I have to import the package `MadNLPMumps` outside the module. Is it right?

```julia
using Pkg
Pkg.activate(@__DIR__)
Pkg.add("ExaModels")
Pkg.add("MadNLPMumps")

module P
    using ExaModels
    function get()
        c = ExaCore()
        x = variable(c, 1)
        objective(c, x[1]^2)
        return ExaModel(c)
    end
end

module A
    using MadNLPMumps
    using ..P
    println("FIRST SOLVE. linear solver: umfpack")
    madnlp(P.get(); print_level=MadNLP.INFO)
    println()
end

using MadNLPMumps # needed outside the module

module B
    using MadNLPMumps
    using ..P
    println("SECOND SOLVE. linear solver: mumps")
    madnlp(P.get(); print_level=MadNLP.INFO)
    println()
end
```

```
FIRST SOLVE. linear solver: umfpack
This is MadNLP version v0.8.12, running with umfpack

Number of nonzeros in constraint Jacobian............:        0
Number of nonzeros in Lagrangian Hessian.............:        1

Total number of variables............................:        1
                     variables with only lower bounds:        0
                variables with lower and upper bounds:        0
                     variables with only upper bounds:        0
Total number of equality constraints.................:        0
Total number of inequality constraints...............:        0
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        0

iter    objective    inf_pr   inf_du inf_compl lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  0.0000000e+00 0.00e+00 0.00e+00 0.00e+00  -1.0 0.00e+00    -  0.00e+00 0.00e+00   0

Number of Iterations....: 0

                                   (scaled)                 (unscaled)
Objective...............:   0.0000000000000000e+00    0.0000000000000000e+00
Dual infeasibility......:   0.0000000000000000e+00    0.0000000000000000e+00
Constraint violation....:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   0.0000000000000000e+00    0.0000000000000000e+00
[...]

PR #510, opened 2025-10-23
[MadNLPMumps] Redefine MadNLP.default_sparse_solver

amontoison:
Related issue: #505 
I will remove `lib/MadNLPMumps` after this PR, I just need a commit to tag a new release of `MadNLPMumps`.


amontoison:
It leads to warning because of the redefinition of `MadNLP.default_sparse_solver`.
I don't want to open a can of worms.
I close the PR.

PR #515, opened 2025-11-24
Add formatter

michel2323:
A format I like that CUDA.jl etc are using. See: https://github.com/fredrikekre/Runic.jl

It will only do suggestions on the PR changes.

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -1,0 +1,100 @@
+name: 'Format'
+
+on:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


frapac:
I am all in to use a formatter, but I don't know how widespread Runic's usage is. What does it bring compared to JuliaFormatter?

michel2323:
I don't know about JuliaFormatter, but Runic lets you apply per commit and can be integrated into git. It's also used for other languages.

PR #519, opened 2025-11-26
Flip objective in stats when maximizing

### LINKED ISSUE #517: Maximization problem and objective value
Hi @amontoison @sshin23 @frapac and others!

If I am not mistaken, when we are maximizing the `stats.objective` value is not the value of the objective but its opposite. If we are considering $\max f(x)$, then at the solution it is returned $-f(x)$. 

In my opinion, the expected behavior is to get $f(x)$ since for instance if we maximize profit we expect to get the profit. What is quite confusing is that if we use JuMP as modeler, then it returns $f(x)$.

My first question. Am I right?
My second. Would it be possible to replace this behavior?

Here is a piece of code:

```julia
using Pkg
Pkg.activate(joinpath(@__DIR__))

using NLPModels
using ExaModels
using MadNLPMumps
using ADNLPModels
using Ipopt
using NLPModelsIpopt
using DataFrames

# min f(x)
f(x) = 1+x[1]^2

function build_problem(problem_type::Symbol, sense::Symbol)
    minimize = sense == :min
    if problem_type == :exa
        c = ExaCore(; minimize=minimize)
        x = variable(c, 1; start=2.0)
        expr = minimize ? f(x) : -f(x)
        objective(c, expr)
        return ExaModel(c)
    elseif problem_type == :adnlp
        objective_function = minimize ? f : (x -> -f(x))
        return ADNLPModel(objective_function, [2.0]; minimize=minimize)
    else
        error("Unsupported problem type: $(problem_type)")
    end
end

function solve_problem(problem, solver_type::Symbol)
    if solver_type == :madnlp
        solver = MadNLPSolver(problem; print_level=MadNLP.ERROR)
        results = solve!(solver)
        println("typeof(results): $(typeof(results))")
        return results.objective
    elseif solver_type == :ipopt
        results = ipopt(problem; print_level=0)
        println("typeof(results): $(typeof(results))")
        return results.objective
    else
        error("Unsupported solver type: $(solver_type)")
    end
end

senses = [:min, :max]
solver_types = [:madnlp, :ipopt]
problem_types = [:exa, :adnlp]

results = DataFrame(;
    sense=String[], solver=String[], problem_type=String[], objective_value=Float64[]
)

for sense in senses
    for solver_type in solver_types
        for problem_type in problem_types
            problem = build_problem(problem_type, sense)
            objective_value = solve_problem(problem, solver_type)
            push!(
                results,
                (
                    sense=String(sense),
                    solver=String(solver_type),
                    problem_type=String(problem_type),
                    objective_value=objective_value,
                ),
            )
        end
    end
end

results
```

We expect to get `1` when minimizing and `-1` when maximizing.

```shell
8×4 DataFrame
 Row │ sense   solver  problem_type  objective_value 
     │ String  String  String        Float64         
─────┼───────────────────────────────────────────────
   1 │ min     madnlp  exa                       1.0
   2 │ min     madnlp  adnlp                     1.0
   3 │ min     ipopt   exa                       1.0
   4 │ min     ipopt   adnlp                     1.0
   5 │ max     madnlp  exa                       1.0
   6 │ max     madnlp  adnlp                     1.0
   7 │ max     ipopt   exa                      -1.0
   8 │ max     ipopt   adnlp                    -1.0
```

klamike:
Fixes #517

### DIFF
--- a/ext/MadNLPMOI/MadNLPMOI.jl
+++ b/ext/MadNLPMOI/MadNLPMOI.jl
@@ -1090,8 +1090,7 @@ MOI.get(model::Optimizer, ::MOI.SolveTimeSec) = model.solve_time
 
 function MOI.get(model::Optimizer, attr::MOI.ObjectiveValue)
     MOI.check_result_index_bounds(model, attr)
-    scale = (model.sense == MOI.MAX_SENSE) ? -1 : 1
-    return scale * model.result.objective
+    return model.result.objective
 end
 
 ### MOI.VariablePrimal
--- a/src/IPM/callbacks.jl
+++ b/src/IPM/callbacks.jl
@@ -3,6 +3,9 @@ function eval_f_wrapper(solver::AbstractMadNLPSolver, x::PrimalVector{T}) where 
     cnt = solver.cnt
     @trace(solver.logger,"Evaluating objective.")
     cnt.eval_function_time += @elapsed begin
+        # NOTE: we flip the objective here so the rest of MadNLP internals
+        # are the same, between min and max. when showing the objective value
+        # to the user (in MadNLPExecutionStats) we flip it back (#517).
         sense = (get_minimize(nlp) ? one(T) : -one(T))
         obj_val = sense * _eval_f_wrapper(solver.cb, variable(x))
     end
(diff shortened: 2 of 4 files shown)


frapac (approved):
LGTM

PR #528, opened 2025-12-15
Move {Sparse,Dense}WrapperModel from MadNLPGPU to MadNLP

frapac:
Moving these two models in MadNLP as they can be useful to wrap on the GPU any model instantiated on the CPU. 

This is what we used to run the CUTEst benchmark on the GPU. 

cc @apozharski


amontoison:
Do we have a note about it somewhere in the documentation?

frapac:
It was not documented before, as this was an unexposed feature in MadNLPTests. We can add a proper documentation in a separate PR.

sshin23 (approved):
LGTM!

PR #529, opened 2025-12-15
Abstract stats

klamike:
@frapac was this intentional? #360 switched it from AbstractMadNLPSolver to MadNLPSolver, but MadIPM was using this

### DIFF
--- a/src/IPM/utils.jl
+++ b/src/IPM/utils.jl
@@ -20,8 +20,9 @@ mutable struct MadNLPExecutionStats{T, VT} <: AbstractExecutionStats
     counters::MadNLPCounters
 end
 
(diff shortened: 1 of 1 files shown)


klamike:
> ERROR: LoadError: could not find source path for package `MbedTLS_jll [c8ffd9c3]` based on /home/action-runner/madnlp-runner-3/_work/MadNLP.jl/MadNLP.jl/docs/Manifest.toml

cache issue? 🤨

frapac:
@klamike thank you for catching this, this was not intentional indeed!

And I confirm we have a cache issue when we build the documentation. This should not block us for merging this PR.

frapac (approved):

PR #533, opened 2025-12-17
[MadNLPGPU] Fix CUDSS uniform batch syntax

klamike:
@amontoison It will need a release after


amontoison:
@klamike Done, sorry for the delay, I was in the plane 🙂

Issue #450, opened 2025-07-11
Passing `x`, `y`, `zl`, and `zu`, to a first call of `solve!` ignores these arguments.


apozharski:
The initialization in [these lines](https://github.com/MadNLP/MadNLP.jl/blob/3e5ceb0e126480bfa2810c255b847d08a2614b8b/src/IPM/solver.jl#L137-L148) is a noop on the first call to `solve!` as the call to `initialize!` later overwrites these changes with the `x0` in `nlp.meta`. It would be nice, however, to be able to pass the initialization (obviously this may have to be pushed away from bounds etc.) without having to modify the `x0` buffer in the nlp as that may lead to subtle side-effects in user code.

frapac:
Solved by #546

PR #546, opened 2026-01-07
Update the signature of the function `solve!` in MadNLP

frapac:
- currently, we can pass the initial variables `x`, `y`, `zl` and `zu` to MadNLP. However, they are all discarded in the function `initialize!`
- this PR removes these arguments, as this behavior is error prone (we should not silently ignore input arguments). 

Solve #450

### DIFF
--- a/src/IPM/solver.jl
+++ b/src/IPM/solver.jl
@@ -124,7 +124,6 @@ function solve!(
     stats::MadNLPExecutionStats;
     kwargs...
 )
-    stats = MadNLPExecutionStats(solver)
     if !isempty(kwargs)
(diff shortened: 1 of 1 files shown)


src/IPM/solver.jl
```diff
@@ -127,27 +119,10 @@ end
 
 # major loops ---------------------------------------------------------
 function solve!(
-    nlp::AbstractNLPModel,
-    solver::AbstractMadNLPSolver,
-    stats::MadNLPExecutionStats;
```
klamike: It would be nice to still have a `solve!` method that allows to pre-allocate the `stats`

PR #547, opened 2026-01-07
Apply `obj_scale` in `unpack_z` in `MakeParameter` case

klamike:
The generic case scales:

https://github.com/MadNLP/MadNLP.jl/blob/ef07381034c80b74a72b4858d587b5f48e914f81/src/Callbacks/nlpmodels.jl#L658-L660

but the MakeParameter one doesn't:

https://github.com/MadNLP/MadNLP.jl/blob/ef07381034c80b74a72b4858d587b5f48e914f81/src/Callbacks/nlpmodels.jl#L689-L692

This PR adds scaling to the latter case. Maybe it should be removed from the former instead?

### DIFF
--- a/src/Callbacks/nlpmodels.jl
+++ b/src/Callbacks/nlpmodels.jl
@@ -688,7 +688,7 @@ function unpack_x!(x_full, cb::SparseCallback{T, VT, VI, NLP, FH}, x) where {T, 
 end
 function unpack_z!(z_full, cb::SparseCallback{T, VT, VI, NLP, FH}, z) where {T, VT, VI, NLP, FH<:MakeParameter}
     free = cb.fixed_handler.free
-    z_full[free] .= z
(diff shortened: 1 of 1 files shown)


klamike:
Note, before the #360 it was unscaled: https://github.com/MadNLP/MadNLP.jl/blob/f96b557d35cc360d7f8141dfe6aeb78296197624/src/IPM/utils.jl#L23-L36

Maybe this PR should be adjusted to removed the scaling from the generic case instead?

klamike:
Perhaps a test can be added that computes the user/un-scaled optimal primal and dual objectives for a convex problem, to check they match?

frapac:
@klamike that's a good suggestion, I will implement this test. We need to be sure the solution returned is correct after post-processing.

frapac:
Your implementation is correct. We should scale the multiplier `z` in the post-processing. The implementation was broken before #360

Issue #537, opened 2025-12-23
Check BFGS and LBFGS (after 🎄 )


amontoison:
I found a bunch of issues with BFGS and LBFGS and optimize many things.
I should do a meeting with François to check that what we have now is correct.

Related PRs: #524 #530 #534 #536

amontoison:
@frapac `MadNLP.solve!` has a specific dispatch if `kkt` is a `SparseKKTSystem` and `QN <: CompactLBFGS` but it doesn't return any error if we still have `QN <: CompactLBFGS` and `kkt` is a `SparseCondensedKKTSystem`.

I didn't understand why I didn't get any error on GPU with LBFGS...

PR #550, opened 2026-01-08
[MOI] Reuse existing data structures from MOI on re-solve

amontoison:
Similar to https://github.com/cvanaret/Uno/pull/474, https://github.com/jump-dev/Ipopt.jl/pull/516, https://github.com/jump-dev/KNITRO.jl/pull/378

### DIFF
--- a/ext/MadNLPMOI/MOI_wrapper.jl
+++ b/ext/MadNLPMOI/MOI_wrapper.jl
@@ -52,6 +52,17 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     nlp_model::Union{Nothing,MOI.Nonlinear.Model}
     ad_backend::MOI.Nonlinear.AbstractAutomaticDifferentiation
     vector_nonlinear_oracle_constraints::Vector{Tuple{MOI.VectorOfVariables,_VectorNonlinearOracleCache}}
+
+    jrows::Vector{Int}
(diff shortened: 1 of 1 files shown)


amontoison:
@franckgaga
We upgraded the MOI interface with François (#548 , #549, #550, #551) .
It could be relevant for you :)

franckgaga:
Niccee, many thanks 🙏 

Is the [VectorNonlinearOracle](https://jump.dev/MathOptInterface.jl/dev/reference/standard_form/#MathOptInterface.VectorNonlinearOracle) also ready for MadNLP?

amontoison:
Yes, we did it in #508.

franckgaga:
Yess, thank you both, will give it a spin for NMPC, I had some trouble with the built-in L-BFGS solver in the past.

amontoison:
I worked a few hours recently on LBFGS and improved the implementation (see #537), I suggest to try it again when the release 0.9 will be available.

@frapac With this PR, I don't recreate the internal "MOI model" (with the expression tree, buffers for AD, coloring structures, etc...) if not needed but I never reuse the workspace of MadNLP.
Compared to Ipopt / KNITRO / Uno, we can also do that and it could be quite nice for @franckgaga.
We just need to use the in-place version of MadNLP (in a different PR).

PR #551, opened 2026-01-08
Support precompilation with MOI extension

amontoison:
It will help for the precompilation.
It is what is used in `NLOpt.jl`, `Ipopt.jl`, `KNITRO.jl` and `UnoSolver.jl`.

Reference: https://github.com/jump-dev/Ipopt.jl/issues/482

### DIFF
--- a/ext/MadNLPMOI/MOI_wrapper.jl
+++ b/ext/MadNLPMOI/MOI_wrapper.jl
@@ -1690,7 +1690,7 @@ end
 
 ### MOI.ConstraintDual
 
-_dual_multiplier(model::Optimizer) = 1.0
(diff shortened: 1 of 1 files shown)


amontoison:
@frapac Is it fine for you?

amontoison:
@frapac You merged your PR before mine and it is a pain to rebase it with all the other ones on top of it.
I opened this one before, be careful please because it is more work for me.

frapac:
> @frapac You merged your PR before mine and it is a pain to rebase it with all the other ones on top of it.
I opened this one before, be careful please because it is more work for me.

We have currently more than 10 open PRs, we cannot ensure we don't break all of them when we merge one. We don't have explicit governance rules, but in general I think it's the responsibility of the PR's author to maintain compatibility with the branch `main`. I also had to do significant rebase in the past.

PR #555, opened 2026-01-11
LBFGS is only supported with the KKT formulation SparseKKTSystem

amontoison:
It is better to have an error then returning the wrong result (it is the case on GPU where we only use the condensed formulation).

### DIFF
--- a/src/IPM/factorization.jl
+++ b/src/IPM/factorization.jl
@@ -161,21 +161,21 @@ function solve!(
     kkt::SparseUnreducedKKTSystem{T, VT, MT, QN},
     w::AbstractKKTVector
     ) where {T, VT, MT, QN<:CompactLBFGS}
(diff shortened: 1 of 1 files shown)


frapac (commented):
LGTM. I would maybe suggest a slightly more explicit error message:
```julia
error("Quasi-Newton approximation of the Hessian is not supported by the KKT formulation ScaledSparseKKTSystem: Please use SparseKKTSystem instead.")
```

amontoison:
> LGTM. I would maybe suggest a slightly more explicit error message:
> 
> ```julia
> error("Quasi-Newton approximation of the Hessian is not supported by the KKT formulation ScaledSparseKKTSystem: Please use SparseKKTSystem instead.")
> ```

Done :+1:

frapac (approved):

PR #557, opened 2026-01-16
[MOI/MadNLPTests] Add `array_type` to support GPU solve from MOI

### LINKED ISSUE #556: [MOI] GPU syntax in readme is not working
It seems the readme syntax for using GPU linear solvers is broken, `Model(()->MadNLP.Optimizer(linear_solver=CUDSSSolver))` tries to init CUDSS based on a CPU sparsecallback. Note the MOI+MadNLPGPU tests are using the wrapper models explicitly.

Perhaps MadNLP should have an API for which array to use given the linear solver type, e.g. `array_type(::Type{CUDSSSolver}) = CuArray`. Then the MOI extension can check if it to see if a wrapper is needed.

klamike:
This PR enables using the syntax 
```julia
using JuMP, MadNLP, MadNLPGPU, CUDA
m = Model(MadNLP.Optimizer)
set_attribute(m, "linear_solver", CUDSSSolver)
set_attribute(m, "array_type", CuArray)
...
optimize!(m)
```
to solve using CUDA/CUDSS via JuMP/MOI. Closes #556

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,11 +111,11 @@ model = Model(()->MadNLP.Optimizer(linear_solver=PardisoMKLSolver))
 ```julia
 using MadNLPGPU, JuMP
 # ...
-model = Model(()->MadNLP.Optimizer(linear_solver=LapackGPUSolver))  # for dense problems
-model = Model(()->MadNLP.Optimizer(linear_solver=CUDSSSolver))      # for sparse problems
+model = Model(()->MadNLP.Optimizer(linear_solver=LapackGPUSolver, array_type=CuArray))  # for dense problems
+model = Model(()->MadNLP.Optimizer(linear_solver=CUDSSSolver, array_type=CuArray))      # for sparse problems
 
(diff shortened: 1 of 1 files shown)


ext/MadNLPMOI/MadNLPMOI.jl
```diff
@@ -1524,8 +1530,22 @@ function MOI.optimize!(model::Optimizer)
         s.eval_hessian_lagrangian_timer = 0.0
     end
     # Instantiate MadNLP.
+    needs_wrap = model.array_type != Vector{Float64} && !(
+        model.nlp isa MadNLP.SparseWrapperModel &&
+        NLPModels.get_x0(model.nlp) isa model.array_type
+    )
```
klamike: This might be too much, I am not entirely clear on when the `nlp` is cleared/reused

klamike:
Docs probably just need a re-run, looks like some network issue

PR #564, opened 2026-01-23
[MOI] Drop support for `ConstraintDualStart` for variable bounds

frapac:


amontoison:
Can you also remove the todo at line 1509?

Issue #525, opened 2025-12-15
[documentation] Explain how to use BFGS in MadNLP


amontoison:
Explain that we have two variants of `BFGS` and the differences with `LBFGS`.
Specify if it works or not on GPU.

Issue #513, opened 2025-11-17
[documentation] Explain how to use an AbstractNLPModels to solve NLP problems on GPU


amontoison:
I should add an NLP model in the documentation implemented without `ExaModels.jl`, to show how to solve generic NLP problems on GPU.
The API is the same on CPU and GPU, example: [here](https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl/blob/main/src/nlp/problems/hs14.jl#L1-L223)

cc @pulsipher

Issue #324, opened 2024-04-12
README.md simplification


sshin23:
Now that MadNLP has mature documentation, it might be better to make README.md minimal and let the user refer to documentation instead. Thoughts? @frapac

Issue #238, opened 2023-01-03
Wrong solution in QuickStart example


haavardhvarnes:
I believe the print of iterations shows wrong results or that the solver does not handle this problem. Objective should end at 306.5 

http://www.stfmc.de/fmc/rhs/x/wschfh.shtml?ampl/tp015.mod
or
https://vanderbei.princeton.edu/ampl/nlmodels/hs/hs015.mod

frapac:
HS015 is a nonconvex problem, with different local solutions. Most interior-point solvers (Ipopt, Knitro, MadNLP) are converging only locally, without guarantee to find the global solution if the problem is non-convex. That's exactly what we are observing in this example. 

The solution with objective 360.4 is also a local solution, albeit not as good as the other local solution with objective 306.5. Usually, one can try to change the initial point in the algorithm to assess if the solver is able to find a different (and potentially better) solution. 

This example is also implemented in Knitro, with the same caveat: https://github.com/jump-dev/KNITRO.jl/blob/master/examples/nlp1.jl#L7-L21

To avoid any confusion in the future, I would suggest modifying the starting point to find the other solution (with objective 306.5) and add a discussion about the local nature of any solution. @haavardhvarnes Do you think that would help to clarify?

frapac:
The documentation has  been updated to explain why HS15 has different local solutions (problem is non-convex). Closing this issue as it has been addressed in #570

PR #570, opened 2026-01-26
Update the documentation for MadNLP 0.9

frapac:
- simplify README (#324 )
- explain how to use BFGS in MadNLP (#525 )
- specify the packages to load for the LBFGS tutorial (#523 )
- explain how to solve an `AbstractNLPModel` on the GPU (#513 )
- explain why we have two solutions for HS15 (#238 )
- improve the documentation for the options without using DocStringExtensions (#85 )
- add a page to detail the algorithm used by MadNLP

### DIFF
--- a/docs/src/man/kkt.md
+++ b/docs/src/man/kkt.md
@@ -20,83 +20,95 @@ the KKT system. The [`AbstractKKTSystem`](@ref) fulfills two goals:
 1. Store the values of the Hessian of the Lagrangian and of the Jacobian.
 2. Assemble the corresponding KKT matrix $$K$$.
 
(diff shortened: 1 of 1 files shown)


amontoison:
--> [preview](https://madsuite.org/MadNLP.jl/previews/PR570/) <--

amontoison:
@frapac You have an error in the first row the KKT system. I also suggest to use `\nabla g(x)` instead of `A` in the first equation of the KKT conditions.
You can use `A` and `W` later and define them from `f` and `g`.
Capture d’écran du 2026-02-01 20-36-00

PR #571, opened 2026-01-28
rename solve! function for KKT and linear systems

frapac:
- for KKT system: `solve!` -> `solve_kkt_system!`
- for linear system: `solve!` -> `solve_linear_system!`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -125,6 +125,9 @@ jobs:
           using Pkg
           Pkg.Registry.update()
           Pkg.develop(path=".")
+          using MadNLP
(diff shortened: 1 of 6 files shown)


amontoison:
@frapac I was reading the documentation today about custom KKT formulations and I am wondering if we should not be consistent for these three routines:
```julia
# Assemble the KKT system
MadNLP.build_kkt!(kkt)

# Factorize the KKT system
MadNLP.factorize!(kkt.linear_solver)

# Backsolve
MadNLP.solve!(kkt, w)
```
and just add the suffix `_kkt!`.

PR #576, opened 2026-02-09
Avoid passing `array_type` to `MadNLPSolver` in MOI

klamike:
Followup to #557. Otherwise we get the misleading warning about ignoring `array_type`:
```julia
julia> set_optimizer_attribute(m, "array_type", CuArray)

julia> optimize!(m)
The following options are ignored: 
 - array_type
This is MadNLP version v0.8.12, running with cuDSS v0.7.1
...
```

Also adds a test that covers the ignoring options path and checks the logs.

### DIFF
--- a/test/madnlp_test.jl
+++ b/test/madnlp_test.jl
@@ -285,7 +285,7 @@ end
 
 @testset "Warn on option ignore" begin
     pipe = Pipe()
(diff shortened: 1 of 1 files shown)


amontoison:
@klamike Do we need the same fix in `MadIPM.jl` ?

PR #587, opened 2026-02-23
Make the sparse wrapper model mutable so the C interface can get pointers to one.

apozharski:
This is necessary for the C interface to be able to wrap CPU models with GPU array types.


frapac:
I am not sure to understand why we need this for the C wrapper. If I look at [the documentation](https://docs.julialang.org/en/v1/manual/types/#Mutable-Composite-Types), a `mutable struct` becomes immutable if we mark all its attributes with `const`: 

> In cases where one or more fields of an otherwise mutable struct is known to be immutable, one can declare these fields as such using const as shown below. This enables some, but not all of the optimizations of immutable structs, and can be used to enforce invariants on the particular fields marked as const.

apozharski:
This is necessary because of a Julia quirk (well more like design decision) with regards to [`pointer_from_objref`](https://docs.julialang.org/en/v1/base/c/#Base.pointer_from_objref). It can only work for heap allocated objects and this seems to be the only way to ensure that the struct is indeed heap allocated. Note `unsafe_pointer_to_objref` does not work as even though it will not throw errors if called on an immutable struct the current "undefined behavior" the corresponding docs describes means that this pointer is occasionally invalidated by Julia optimizations which leads to Segmentation Faults.

apozharski:
I have chosen to wrap the structs in `Ref` so this is no longer necessary.

PR #589, opened 2026-02-23
Report the number of KKT factorizations in MadNLP

frapac:


amontoison:
I like the content of this PR!

PR #598, opened 2026-03-09
Add SchurComplementKKTSystem for two-stage stochastic programs

michel2323:
I need some feedback here: @frapac @sshin23 .

First stab at  Schur complement decomposition for block-arrowhead KKT systems arising from two-stage stochastic NLPs modeled with ExaModels (https://github.com/exanauts/ExaModels.jl/pull/244).

* Condensed formulation
* Added `linear_solver` selects solver for S, not the blocks
* The CPU and GPU Schur KKT systems are different because CPU and GPU parallel implementations are quite different. One may change that, but I found this option cleaner
* Inertia in both cases is:  `is_inertia_correct()`: S (not full system) is SPD → `num_neg == 0`
* Mapping and data transfer:
 - CPU: `ScenarioBlockMap` precomputes all COO→nzval index mappings at construction (zero runtime searching), Per-scenario loops parallelized with @blas_safe_threads
 - GPU: 12 KernelAbstractions kernels for Hessian/Jacobian scatter, diagonal injection, inequality condensing, RHS extract/writeback, batched CUDSS, CUBLAS gemm batched accumulation, CUBLAS dense solver

Not yet done:
* Auto-detection of `ns/nv/nd/nc` from `TwoStageTags` (WIP)


michel2323:
I've deduplicated a lot of code and tried to simplify. @frapac, this is ready to take a look now. The most complex bit is the constructor.

docs/src/tutorials/schur.md
```diff
@@ -0,0 +1,211 @@
...
+    lvar_v = fill(-100.0, nv, ns), uvar_v = fill(100.0, nv, ns),
+    lvar_d = fill(-100.0, nd),     uvar_d = fill(100.0, nd),
+)
+nothing
+```
+
+## Solving with `SchurComplementKKTSystem`
+
+[`SchurComplementKKTSystem`](@ref) needs to know the four scenario dimensions
+`(ns, nv, nd, nc)` so it can carve the global sparsity into per-scenario
```
frapac: shouldn't the size of all scenarios be the same?

docs/src/tutorials/schur.md
```diff
@@ -0,0 +1,211 @@
...
+    The Schur system reduces the linear-algebra cost from one factorization of
+    a single sparse system of size ``n_s(n_v + n_c) + n_d`` to ``n_s``
+    independent factorizations of size ``n_v + n_c^{eq}`` plus one dense
+    factorization of size ``n_d``. The per-scenario factorizations are
+    independent, which is what makes the GPU variant a batched call. The
+    payoff grows with `ns` and shrinks with `nd`.
+
+### Choosing the per-scenario solver
+
+By default each per-scenario block is factored by `MadNLP.LDLSolver`. On the
```
frapac: Why not using `MumpsSolver` by default? I suspect the per-scenario matrix to remain symmetric indefinite, and LDLSolver requires a SQD matrix.

docs/src/tutorials/schur.md
```diff
@@ -0,0 +1,211 @@
...
+[`SchurComplementKKTSystem`](@ref) can recover `(ns, nv, nd, nc)` directly from
+them. The convention is `tags.var_scenario[i] == 0` for design variables,
+`tags.var_scenario[i] == k` (with `k ∈ 1:ns`) for scenario `k` recourse
+variables, and the same encoding for `tags.con_scenario`. When the tags are
+present, you can omit `kkt_options` entirely:
+
+```julia
+using ExaModels
+
+core = ExaModels.TwoStageExaCore(Float64)
```
frapac: Should we assume `TwoStageExaCore` will stay in ExaModels?

docs/src/tutorials/schur.md
```diff
@@ -0,0 +1,211 @@
...
+end
+ExaModels.objective(core, (d[j] - 1.0)^2 for j in 1:nd)
+nlp = ExaModels.ExaModel(core)
+
+results = madnlp(nlp; kkt_system = SchurComplementKKTSystem)
+```
+
+## Variable and constraint ordering
+
+The Schur reduction relies on a strict block layout — `madnlp` will throw an
```
frapac: How do we detect a mispecified variable layout?

src/KKT/Schur/schur.jl
```diff
@@ -0,0 +1,1006 @@
...
+    ineq_per_scenario = Vector{Vector{Int}}(undef, ns)
+
+    nc_eq_per_s = 0
+    nc_ineq_per_s = 0
+    for k in 1:ns
+        cr = (k-1)*nc+1 : k*nc
+        eq_per_scenario[k] = Int[]
+        ineq_per_scenario[k] = Int[]
+        for gi in cr
+            if gi in ind_eq_set
```
frapac, line 310: this can be expensive, as we have three for look imbricated here.

src/KKT/Schur/schur.jl
```diff
@@ -0,0 +1,1006 @@
...
+            for (coo_a, da) in d_entries, (coo_b, db) in d_entries
+                push!(ineq_S_row_vec, da)
+                push!(ineq_S_col_vec, db)
+                push!(ineq_S_jcoo1_vec, coo_a)
+                push!(ineq_S_jcoo2_vec, coo_b)
+                push!(ineq_S_bufidx_vec, bidx)
+            end
+        end
+
+        block_maps[k] = ScenarioBlockMap(
```
frapac, line 635: Introducing a structure `ScenarioBlockMap` is a good idea

src/KKT/Schur/schur.jl
```diff
@@ -0,0 +1,1006 @@
...
+    block_maps = sym.block_maps
+    eq_per_scenario = sym.eq_per_scenario
+    ineq_per_scenario = sym.ineq_per_scenario
+
+    # Per-scenario A_kk: independent copies of the shared template (same sparsity, fresh nzval).
+    A_kk_vec = [copy(sym.akk_csc_template) for _ in 1:ns]
+
+    # --- Dense matrices ---
+    aug_com = Matrix{T}(undef, nd, nd)
+    C_dk = [Matrix{T}(undef, nd, blk_size) for _ in 1:ns]
```
frapac, line 837: wouldn't it be more appropriate to store `C_dk` in sparse format?

src/KKT/Schur/schur.jl
```diff
@@ -0,0 +1,1006 @@
...
+    # Initialize Schur complement S = H_dd + diag(pr_diag_dd)
+    S = kkt.aug_com
+    fill!(S, zero(T))
+    _scatter_add!(S, kkt.hess, kkt.hess_S_coo, kkt.hess_S_row, kkt.hess_S_col)
+    @inbounds for i in 1:nd
+        S[i, i] += kkt.pr_diag[ns*nv+i]
+    end
+
+    # Phase 1 (parallel): assemble per-scenario blocks, factorize, compute A_kk^{-1} * C_dk'
+    @blas_safe_threads for k in 1:ns
```
frapac, line 953: Maybe we should allow running this for loop in parallel on the CPU using multithreading

frapac (commented):
This is my preliminary comments. The structure looks good overall. I will try to run the code on my computer to understand it better. 

Two questions:
- would it make sense to run the assembling and the factorization of the KKT system in parallel on the CPU as well, using multithreading?
- should we move this in a new package or keep it in MadNLP?

michel2323:
> * would it make sense to run the assembling and the factorization of the KKT system in parallel on the CPU as well, using multithreading?

Currently, the factorization is already done in parallel using Julia threads over the scenarios using `@blas_safe_threads` defined in https://github.com/MadNLP/MadNLP.jl/blob/88a6dde45ba3fff0ef712da0a2ee7875f7511744/src/utils.jl#L99-L106 .

However, I am open to any suggestions (nested threaded, MPI, etc).

> * should we move this in a new package or keep it in MadNLP?

I've opened an issue to track this discussion. From what I remember, the preliminary conclusion in the last meeting was to include batched and two-stage KKT (Schur complement with the same number of scenarios) in the base MadNLP. This is also related to the discussed install-friction problem (more packages to add, version compat to manage).

docs/src/tutorials/schur.md
michel2323: Yes. We want to keep a uniform per-scenario shape for the two-stage implementation in MadNLP. I think a non-uniform shape would warrant a new package, such as MadNLPStochastic.jl. This would also break the GPU SIMD parallelism and be a drastically different implementation.

I will add this explicitly to the tutorial. Currently, the solver could fail silently with wrong results if `n == ns·nv + nd` and `m == ns·nc` and non-uniform sparsity. I'll add graceful failure and more than just asserts.

docs/src/tutorials/schur.md
michel2323: Good point. How about `UmfpackSolver`, or why have you picked Mumps as the default in base MadNLP?

docs/src/tutorials/schur.md
michel2323: I am not sure @sshin23, but something similar. Could we just remove this tutorial from the first PR? It's cumbersome to define a two-stage model without a modeling framework.

docs/src/tutorials/schur.md
michel2323: See above the same-sized scenario comment. This should be caught and non-uniform sparsity pattern is missed at the moment.

src/KKT/Schur/schur.jl
michel2323: Yes. However, this is a one-time cost. Given that this implementation is a single-node one, I'd argue this should not be an issue, as the system can never get bigger than the memory available on a single node. But I guess one could vectorize it.

src/KKT/Schur/schur.jl
michel2323: Right. This would require much more code. I am not sure whether to include it in this PR which is already quite large, or in a follow-up.

src/KKT/Schur/schur.jl
michel2323: See above. That's what `@blas_safe_threads` does. Maybe a bad name.

frapac (approved):
LGTM!

PR #603, opened 2026-03-23
Removing 99% of allocations in `solve!`

apozharski:
Currently the use of `mapreduce` everywhere causes a relatively large amount of allocations to happen during `solve!`. This is my initial attempt at fixing these, though in all likelihood I have broken some GPU functionality which was relying on the multiple dispatch of `mapreduce`.

### DIFF
--- a/lib/MadNLPGPU/src/IPM/kernels.jl
+++ b/lib/MadNLPGPU/src/IPM/kernels.jl
@@ -452,3 +452,7 @@ function get_rel_search_norm(x::AbstractGPUVectorOrSubVector{T}, dx::AbstractGPU
         x, dx
     )
 end
+
+function populate_RR_nn!(nn, c, mu, rho)
+    map!((c) -> (mu - rho*c)/(2*rho)+sqrt(((mu-rho*c)/(2*rho))^2 + mu*c/(2*rho)))
+end
--- a/lib/MadNLPGPU/src/IPM/kernels.jl
+++ b/lib/MadNLPGPU/src/IPM/kernels.jl
@@ -453,6 +453,6 @@ function get_rel_search_norm(x::AbstractGPUVectorOrSubVector{T}, dx::AbstractGPU
     )
 end
 
-function populate_RR_nn!(nn, c, mu, rho)
+function populate_RR_nn!(nn::AbstractGPUVectorOrSubVector{T}, c::AbstractGPUVectorOrSubVector{T}, mu, rho) where T
     map!((c) -> (mu - rho*c)/(2*rho)+sqrt(((mu-rho*c)/(2*rho))^2 + mu*c/(2*rho)))
 end
--- a/lib/MadNLPGPU/src/IPM/kernels.jl
+++ b/lib/MadNLPGPU/src/IPM/kernels.jl
@@ -1,0 +1,439 @@
+# TODO(@anton): All of these could probably be made more efficient via custom kernels.
+#               For now I am just returning the mapreduce functionality here so MadNLPGPU works again.
+
+function get_varphi(obj_val, x_lr::VT, xl_r::VT, xu_r::VT, x_ur::VT, mu) where {VT <: AbstractGPUVector}
+    return obj_val + mapreduce(
+        (x1,x2) -> _get_varphi(x1,x2,mu), +, x_lr, xl_r
+    ) + mapreduce(
+        (x1,x2) -> _get_varphi(x1,x2,mu), +, xu_r, x_ur
+    )
+end
+
+function get_inf_du(f::VT, zl::VT, zu::VT, jacl::VT, sd) where {VT <: AbstractGPUVector}
+    return mapreduce((f,zl,zu,jacl) -> abs(f-zl+zu+jacl), max, f, zl, zu, jacl; init = zero(eltype(f))) / sd
+end
+
+
+function get_inf_compl(x_lr::VT, xl_r::VT, zl_r::VT, xu_r::VT, x_ur::VT, zu_r::VT, mu, sc) where {VT <: AbstractGPUVector}
+    return max(
+        mapreduce(
+            (x_lr, xl_r, zl_r) -> abs((x_lr-xl_r)*zl_r-mu),
+            max,
+            x_lr, xl_r, zl_r;
+            init = zero(eltype(x_lr))
+        ),
+        mapreduce(
+            (xu_r, x_ur, zu_r) -> abs((xu_r-x_ur)*zu_r-mu),
+            max,
+            xu_r, x_ur, zu_r;
+            init = zero(eltype(x_lr))
+        )
+    ) / sc
+end
+
+function get_min_complementarity(x_lr::AbstractGPUVector{T}, xl_r::AbstractGPUVector{T}, zl_r::AbstractGPUVector{T},
+                                 x_ur::AbstractGPUVector{T}, xu_r::AbstractGPUVector{T}, zu_r::AbstractGPUVector{T}) where T
+    cc_lb = mapreduce((x_l, xl, zl) -> (x_l-xl)*zl, min, x_lr, xl_r, zl_r, init=T(Inf))
+    cc_ub = mapreduce((x_u, xu, zu) -> (xu-x_u)*zu, min, x_ur, xu_r, zu_r, init=T(Inf))
+    return min(cc_lb,cc_ub)
+end
+
+function get_varphi_d(
+    f::AbstractGPUVector{T},
+    x::AbstractGPUVector{T},
+    xl::AbstractGPUVector{T},
+    xu::AbstractGPUVector{T},
+    dx::AbstractGPUVector{T},
+    mu,
+) where T
+    return mapreduce(
+        (f,x,xl,xu,dx)-> (f - mu/(x-xl) + mu/(xu-x)) * dx,
+        +,
+        f, x, xl, xu, dx;
+        init = zero(T)
+    )
+end
+
+function get_alpha_max(
+    x::AbstractGPUVector{T},
+    xl::AbstractGPUVector{T},
+    xu::AbstractGPUVector{T},
+    dx::AbstractGPUVector{T},
+    tau,
+) where T
+    return min(
+        mapreduce(
+            (x, xl, dx) -> dx < 0 ? (-x+xl)*tau/dx : T(Inf),
+            min,
+
+            x, xl, dx,
+            init = one(T)
+        ),
+        mapreduce(
+            (x, xu, dx) -> dx > 0 ? (-x+xu)*tau/dx : T(Inf),
+            min,
+            x, xu, dx,
+            init = one(T)
+        )
+    )
+end
+
+function get_alpha_z(
+    zl_r::AbstractGPUVector{T},
+    zu_r::AbstractGPUVector{T},
+    dzl::AbstractGPUVector{T},
+    dzu::AbstractGPUVector{T},
+    tau,
+)  where T
+    return min(
+        mapreduce(
+            (zl_r, dzl) -> dzl < 0 ? (-zl_r)*tau/dzl : T(Inf),
+            min,
+            zl_r, dzl,
+            init = one(T)
+        ),
+        mapreduce(
+            (zu_r, dzu) -> dzu < 0 ? (-zu_r)*tau/dzu : T(Inf),
+            min,
+            zu_r, dzu,
+            init = one(T)
+        )
+    )
+end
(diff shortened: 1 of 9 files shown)


klamike:
Some related work: https://github.com/klamike/MadIPM.jl/blob/mk/batchperf/ext/MadIPMCUDAExt/cuda_mapreduce.jl

apozharski:
GPU support is now unbroken by providing `AbstractGPUArray` and co specializations which revert to the old approach. This is obviously a bit of an unsatisfying "fix" and I am happy to take a look at this later, but selfishly I don't really have too much time currently :sweat_smile:.

This should be ready for review now @frapac

apozharski:
> Some related work: https://github.com/klamike/MadIPM.jl/blob/mk/batchperf/ext/MadIPMCUDAExt/cuda_mapreduce.jl

Yea I think something like this would be good to replace the `mapreduce` use in `MadNLPGPU` though I have seen rumors that the `mapreduce` implementation for `CuVector` doesn't allocate even in the multi-array case.

klamike:
It depends on how many blocks it decides to use: https://github.com/JuliaGPU/CUDA.jl/blob/b4ac08a7b4722331f0bcb5c0defa60d1c22566f9/src/mapreduce.jl#L243-L292

The atomics perf seemed fine (only profiled on newer blackwell GPU though), an alternative would be to pre-allocate some workspace for the partials.

frapac (commented):
I have only minor comments so far. My main concern is that we should not introduce a performance regression when updating the IPM kernels

src/IPM/kernels.jl
```diff
@@ -339,12 +347,11 @@ function get_varphi_d(
...
     mu,
 ) where T
-    return mapreduce(
-        (f,x,xl,xu,dx)-> (f - mu/(x-xl) + mu/(xu-x)) * dx,
-        +,
-        f, x, xl, xu, dx;
-        init = zero(T)
-    )
+    varphi_d = zero(T)
+    @simd for ii in eachindex(f)
```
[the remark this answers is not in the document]
apozharski: I will look briefly at the LLVM and native code produced for this with and without `@simd` but from looking at the set of LLVM optimization passes the compiler takes I think you are right.

apozharski:
> I have only minor comments so far. My main concern is that we should not introduce a performance regression when updating the IPM kernels

Yep, agreed with that. When testing with `CCOpt` on MacMPEC I see a minor (<2% avg) performance improvement when comparing using the `disable_garbage_collector` flag (including https://github.com/JuliaSmoothOptimizers/AmplNLReader.jl/pull/168 this PR for both main and this branch), but it would be good to run a comparison between main and this branch on some test suite (I guess CUTEst?). ~~Is there a standard test you run to compare?~~ Nevermind, I should just look for 2 seconds before I ask :)

apozharski:
It seems to me that the loops are _much_ faster even discounting garbage collector overhead:
The old `mapreduce`:
```julia
julia> @benchmark MadNLP.get_inf_du(f, zl, zu, jacl, sd) setup=(f=rand(10000); zl=rand(10000); zu=rand(10000); jacl=rand(10000); sd=rand();GC.enable(true);GC.gc();GC.enable(false))
BenchmarkTools.Trial: 27 samples with 1 evaluation per sample.
 Range (min … max):  25.020 μs … 37.888 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     26.637 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   27.442 μs ±  2.473 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▁█▁█  ▄            ▁                                     
  ▆▁▆▆████▆▆█▆▁▁▆▁▆▁▁▁▁▆▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆ ▁
  25 μs           Histogram: frequency by time        37.9 μs <

 Memory estimate: 78.20 KiB, allocs estimate: 3.
```

and the new loops
```julia
julia> @benchmark MadNLP.get_inf_du(f, zl, zu, jacl, sd) setup=(f=rand(10000); zl=rand(10000); zu=rand(10000); jacl=rand(10000); sd=rand();GC.enable(true);GC.gc();GC.enable(false))
BenchmarkTools.Trial: 27 samples with 8 evaluations per sample.
 Range (min … max):  4.708 μs …   7.905 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     5.067 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.629 μs ± 991.219 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    █                                                          
  ▄▄█▄▄▆▆▁▁▁▁▁▁▄▁▁▁▁▁▆▄▁▁▁▁▄▁▁▁▁▄▁▁▁▄▁▁▁▁▄▁▁▁▄▁▁▁▄▁▁▁▁▁▁▄▁▁▁▄ ▁
  4.71 μs         Histogram: frequency by time         7.9 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

klamike:
may want to do `reduce_rhs!` and `finish_aug_solve!` as well, at least in the batch case on GPU they seem to allocate (though it is broadcasting, not mapreduce)

apozharski:
> may want to do `reduce_rhs!` and `finish_aug_solve!` as well, at least in the batch case on GPU they seem to allocate (though it is broadcasting, not mapreduce)

image

(Captured from a CCOpt test because I have the infrastructure set up for that already)

It seems on CPU the compiler is smart enough to do the broadcasts without an intermediate vector.

apozharski:
Caught a few more allocations in the restoration phase code, as well as a type instability caused by `RobustRestorer`.

There are still a few type instabilities caused by the `AbstractBarrierUpdate` type in `MadNLPOptions` but the few ideas i've tried to get rid of this are ugly unfortunately. Made an issue #604.

apozharski:
The last commit fixes a bit of a spooky bug in `get_obj_val_RR` which was only taking into account the objective of the first `min(nvar, ncon)` entries in the constraints or variables (found thanks to @frapac  running this PR against the CUTEst benchmark). We maybe backport this into a bugfix release.

apozharski:
I think this PR will need a rebase after #497 is merged, and I will do a brief test to make sure that there are no regressions from that. After that I think all that is left is review and merge.

apozharski:
Also to document what still allocates (re: @sshin23  your question during the development meeting).

1. An extra jacobian buffer is needed during initialization.
2. The filter allocates when it expands
3. `@sprintf` allocates when logging is turned on.

These are the major remaining places I see.

frapac:
@klamike we need your input with this PR

lib/MadNLPGPU/src/IPM/kernels.jl
```diff
@@ -0,0 +1,464 @@
+# TODO(@anton): All of these could probably be made more efficient via custom kernels.
+#               For now I am just returning the mapreduce functionality here so MadNLPGPU works again.
+
```
klamike: consider removing before merge
```suggestion
```

src/IPM/kernels.jl
```diff
@@ -270,34 +270,37 @@ function _get_varphi(x1::T, x2::T, mu::T) where T
 end
 
 function get_varphi(obj_val, x_lr, xl_r, xu_r, x_ur, mu)
-    return obj_val + mapreduce(
-        (x1,x2) -> _get_varphi(x1,x2,mu), +, x_lr, xl_r
-    ) + mapreduce(
-        (x1,x2) -> _get_varphi(x1,x2,mu), +, xu_r, x_ur
-    )
+    varphi = obj_val
+    # TODO(@anton) Check if we can @simd this
```
klamike:
```suggestion
```

klamike (approved):
LGTM! just a couple comments that can be removed.

We may want to add a non-allocating `mapreduce` for GPU in a follow-up, it helped in MadIPM-Batch and isn't too complicated. I think that would be better/easier to maintain than writing custom kernels for each function.

lib/MadNLPGPU/src/IPM/kernels.jl
apozharski: I think this one we keep until
>We may want to add a non-allocating mapreduce for GPU in a follow-up

is handled, just as a reminder (perhaps without the `TODO`)

apozharski:
> LGTM! just a couple comments that can be removed.
> 
> We may want to add a non-allocating `mapreduce` for GPU in a follow-up, it helped in MadIPM-Batch and isn't too complicated. I think that would be better/easier to maintain than writing custom kernels for each function.

Thanks Michael! I guess in the GPU case the choice of allocating vs non-allocating versions is a bit more fraught though right? Because there is the need to balance saturating the GPU vs the performance penalty of allocating intermediate buffers. 

I will deal with the comments and rebase this branch hopefully today, then it should be good to go in.

frapac (approved):

frapac:
The tests are failing on the GPU runners, but it is not related to this PR: the NVIDIA V100 we are using do not support CUDA 13. 

We have an ongoing discussion with @michel2323 on how to adapt our CI.

michel2323:
> The tests are failing on the GPU runners, but it is not related to this PR: the NVIDIA V100 we are using do not support CUDA 13.
> 

If merging into master, it will be resolved by https://github.com/MadNLP/MadNLP.jl/pull/615 . We are just doing another `v0.9.2` release before merging and moving from `CUDA` to `CUDACore` across the stack. You can also just rebase onto that branch. Note that this should work eventually on a V100 once this is released https://github.com/JuliaGPU/CUDA.jl/pull/3134 .

apozharski:
> > The tests are failing on the GPU runners, but it is not related to this PR: the NVIDIA V100 we are using do not support CUDA 13.
> 
> If merging into master, it will be resolved by #615 . We are just doing another `v0.9.2` release before merging and moving from `CUDA` to `CUDACore` across the stack. You can also just rebase onto that branch. Note that this should work eventually on a V100 once this is released [JuliaGPU/CUDA.jl#3134](https://github.com/JuliaGPU/CUDA.jl/pull/3134) .

Thanks for letting me know @michel2323 and @frapac. I had planned to dig in to what it was I broke this afternoon but from what you said and the fact that running the gpu tests on my machine seem to pass I guess we either wait for the GPU work to merge and rebase this or merge this accepting a red main for now.

src/IPM/utils.jl
```diff
@@ -104,11 +109,17 @@ function get_cons_info(solver)
...
+    num_ineq_cons = length(solver.cb.ind_ineq)
+    num_le_cons = 0
+    num_ue_cons = 0
+    for ii in eachindex(g_lb)
+        if (g_lb != -Inf) && (g_ub ==  Inf)
+            num_le_cons += 1
+        elseif (g_ub !=  Inf) && (g_lb == -Inf)
+            num_ue_cons += 1
+        end
+    end
```
cstjean: I'm just a bystander randomly reviewing this, but this look off? You're not using `ii` anywhere inside the loop. I guess you meant `g_lb[ii]`?

The loop works but FYI `using LazyArrays`, this would be:
```suggestion
    num_le_cons = sum(@~ (g_lb .!= -Inf) .& (g_ub .==  Inf))
    num_ue_cons = sum(@~ (g_ub .!=  Inf) .& (g_lb .== -Inf))
```
which is non-allocating. There's also 
```julia
imap = Base.Iterators.map
num_le_cons = sum(imap(&, imap(!=(-Inf), g_lb), imap(==(Inf), g_ub)))
```
though it's not that easy to read.

The other loop above has the same problem (`ii` isn't used)

apozharski: Oh thanks for reminding me about this! I noticed this bug in the logs of some problems but fixing it never made it onto my todo list.

`LazyArrays` is indeed cool but I personally find the `@~` syntax makes my brain slam on the brakes when it encounters it, similarly to the iterator based vectorization. I always prefer the raw loops, though perhaps that is due to C being being my comfort language.

michel2323:
@apozharski Can you rebase? The CI should pass then and we merge.

apozharski:
> @apozharski Can you rebase? The CI should pass then and we merge.

Rebased, I didn't run an alloc profile yet but just looking at the changes since the last rebase I think this still removes all the allocations minus the ones I mentioned previously.

apozharski:
Argh the last bugfix broke gpu. 

Will fix when I get home (by adding a specialization for GPUArrays :smiling_face_with_tear: )

PR #614, opened 2026-05-09
[MadNLPGPU] Bump to CUDA.jl v6 (subpackages) and CUDSS 0.7

### LINKED ISSUE #610: Up CUDA to v6
CUDA refactored to improve startup time. Major number is upped for precaution, but the upgrade should not be breaking in any way.

KSepetanc:
Closes #610.

Marking this as draft until CUDSS 0.7 is released ([CUDSS PR #144](https://github.com/exanauts/CUDSS.jl/pull/144) is already open for it). CUDSS will use subpackages instead of CUDA meta package.

My take is that we need to mirror that here as it allows users to load only minimum required dependencies, but it is still non breaking, i.e. it will work even if someone still loads with `using CUDA` as CUDA.jl is superset.


KSepetanc:
Let me know if you intend to bump version to 0.9 or 0.8.1 so I can set downstream.

0.8.1 is defensible as no breakage is caused by this (`using CUDA` still works).

michel2323:
Took commit to https://github.com/MadNLP/MadNLP.jl/pull/615

KSepetanc:
@michel2323 , prepared the whole stack for upgrade (HybridKKT, MadNLPIPM, MadNCL).

PR #615, opened 2026-05-11
compat CUDA@6.0.0 CUDSS@0.7

michel2323:
Bumped MadNLPGPU from 0.8 to 0.9 because of CUSPARSE -> cuSPARSE name changes.


michel2323:
I'm not sure ExaModels (v0.9 release is not CUDA 6 compatible) should be a dependency for the MadNLP CI, just as we discussed for the Schur KKT implementation.

Next ExaModels release won't have CUDA as a dependency though.

michel2323:
@sshin23 The CUDA runners are based on V100 GPUs, which are not compatible with CUDA 13. However, CUDA.jl 6.0 requires CUDA 13. We can use kkt at ANL if you want.

michel2323:
I think we should go straight for MadNLP 0.10, ExaModels 0.10, CUDSS.jl 0.7, and CUDA.jl 6.0.

frapac (approved):

KSepetanc:
Please check what is going on with registry tagger. There is still latest 0.8.0, so current and last release are missing.

PR #618, opened 2026-05-22
Release 0.10.0

michel2323:
### MadNLP v0.10.0

> Highlights: the main `solve!` loop is now allocation-free in steady state — ~99% of per-iteration allocations have been removed by replacing `mapreduce`-based reductions with explicit loops and extracting backend-specific kernels so that GPU code paths can plug in without touching the core. Plus a CUDA.jl v6 / CUDSS 0.7 / Julia 1.12 update for `MadNLPGPU`.

**Performance**
- Remove ~99% of per-iteration allocations from `solve!`. `get_varphi`, `get_inf_du`, `get_inf_compl`, `get_min_complementarity`, and friends in `src/IPM/kernels.jl` are rewritten as explicit loops instead of `mapreduce`. `count_lu_bounds` replaces the boolean-mask + `sum` pattern in `get_vars_info` / `get_cons_info`. Indexed updates in `_set_aug_diagonal!` are wrapped in `@views` to avoid temporary copies. The unused dense `jac_buffer` allocation in the NLPModels callback is removed. (#603)
- New `populate_RR_nn!` kernel for restoration-phase initialization, factored out so backends can override it. (#603)

**API changes**
- `RR` field on `MadNLPSolver` is now typed `Union{Nothing,RobustRestorer{T,VT}}` (was `RobustRestorer{T}`). Downstream code that holds a typed reference to `RobustRestorer` needs the extra type parameter. (#603)
- `get_vars_info` and `get_cons_info` are now restricted to `AbstractMadNLPSolver`. (#603)
- New extension points (`_get_varphi`, `get_varphi`, `get_inf_du`, `get_inf_compl`, `get_min_complementarity`, `get_varphi_d`, `get_alpha_max`, `get_alpha_z`, restoration-phase `get_*_R` variants, `populate_RR_nn!`, `count_lu_bounds`, `SubVector`) are exported from `MadNLP` for backends to specialize. (#603)

---

### MadNLPGPU v0.10.0

> Highlights: bumped to CUDA.jl v6 and CUDSS 0.7, and GPU specializations of the new MadNLP core kernels.

**Dependencies**
- Require CUDA.jl ≥ 6.0.0 and CUDSS ≥ 0.7. (#615)
- Depend on the CUDA subpackages (`CUDACore`, `cuBLAS`, `cuSOLVER`, `cuSPARSE`) instead of the `CUDA` meta-package. The CUDA extension now activates from these subpackages directly, so MadNLPGPU works alongside any CUDA.jl v6.x without forcing the meta-package into the project. (#615)
- AMDGPU.jl ≥ 2.

**New features**
- GPU specializations of the IPM kernels refactored out in MadNLP #603 (`get_varphi`, `get_inf_du`, `get_inf_compl`, `get_min_complementarity`, restoration `populate_RR_nn!`, etc.), dispatched on `AbstractGPUVectorOrSubVector`. The GPU path keeps using `mapreduce` while the CPU path uses explicit loops — both avoid the per-iteration allocations that the prior code had. (#603)

**Fixes**
- Work around a Julia 1.12 regression with the global solver state in `MadNLPGPU.jl`. (#615)

**Compatibility**
- Requires MadNLP ≥ 0.10.0.

---


michel2323:
MadNLPHSL needs a bump too

---

### MadNLPHSL v0.7.2

- Bump compat to MadNLP ≥ 0.10.

michel2323:
ExaModels needs a compat bump for each minor MadNLP bump. What about MadNLP 1.x?

apozharski:
I'd like to finally update `libMad` dependencies to tagged versions (and stop having to maintain backports of AOT fixes in `CUDA.jl`), so is there something blocking this new tag from going in?

Regarding:
> ExaModels needs a compat bump for each minor MadNLP bump. What about MadNLP 1.x?

I think this is just a matter of "doing [Julia] SemVer correctly", unfortunately, as this a "breaking" set of changes for some packages, as it changes the set of exported functions (possibly breaking peoples' code). If we were to be in the 1.x.x regime we would have to bump to 2.0.0 in this case causing the same pain anyway.

Though, I wonder if the somewhat circular dependency of ExaModels <-> MadNLP in the docs could be rectified somehow because this will make every breaking change a bit annoying.

michel2323:
We have actually now removed the MadNLP dependency from ExaModels in v0.11. Fingers crossed...

PR #621, opened 2026-06-04
[MadNLPGPU] Support CUDSS 0.8

michel2323:


lib/MadNLPGPU/Project.toml
```diff
@@ -30,7 +30,7 @@ AMD = "0.5"
 AMDGPU = "2"
 Adapt = "4"
 CUDACore = "6"
-CUDSS = "0.7"
+CUDSS = "0.8"
```
frapac: genuine question: can we support both cuDSS.jl 0.7 and 0.8?

amontoison:
You also need to update https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl#L2-L78

lib/MadNLPGPU/Project.toml
michel2323: Good point. We can.

michel2323: 0.7 is kept. I think that's a good sanity check and we can deprecate it later.

michel2323:
> You also need to update https://github.com/MadNLP/MadNLP.jl/blob/master/lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl#L2-L78

This should be fixed now.

PR #627, opened 2026-07-01
Store variable & constraint names in the MOI wrapper

guimarqu:
@frapac To build a tool that identifies degenerate nonlinear programs, we need to be able to associate the constraint & variable names from the model with the internal col & row ids from MadNLP.

This PR adds attributes to the `Optimizer` struct in order to store the variable & constraints together with their associated MOI variable or constraint index. 

The attributes are mappings :
- from the variable index to the variable name
- from the constraint index to the constraint name
- from the variable name to the variable index
- from the constraint name to the constraint index

This is still a draft PR because I'm not sure whether the last two mappings are useful.


amontoison:
cc @blegat 
Benoît, may I ask you to review the modifications?

blegat (approved):
Looks good to me :)

PR #629, opened 2026-07-14
Remove hard dependency on `MadNLPTests` in `MadNLPGPU` and `MadNLPHSL`

apozharski:
This reverts changes added in [`f99a4ad`](https://github.com/MadNLP/MadNLP.jl/commit/f99a4ad105107d60751155ddcd9692fea3740e94) which would now cause `libMad` to pull in and ship the entirety of `JuMP` as a transitive dependency of `MadNLPTests`

### DIFF
--- a/lib/MadNLPGPU/Project.toml
+++ b/lib/MadNLPGPU/Project.toml
@@ -9,7 +9,6 @@ GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MadNLP = "2621e9c9-9eb4-46b1-8089-e8c72242dfb6"
-MadNLPTests = "b52a2a03-04ab-4a5f-9698-6a2deff93217"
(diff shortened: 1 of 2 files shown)


apozharski:
`test-hsl` error is unrelated and caused by something being broken with the registry it seems.

apozharski:
Could someone with permissions try re-running the amdgpu test. The environment is identical to the previous commit, so I believe this test is flakey.

amontoison:
Done

apozharski:
Oh btw, because the `runic` workflow can be filtered when no `*.jl` files are touched it shouldn't be a required check ([see this note in the github actions docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-paths)) . It should be disabled as a required check imo or it should always be run

amontoison (approved):

amontoison:
@apozharski Can you open a PR for that?
I will merge this PR with the bypass rules and register the releases.

apozharski:
Yea let me think about the best way to do that, because either you can drop it from the required checks by changing the protection ruleset for master (which is in the repo settings and I don't have permissions for :sweat_smile: ) or I can change the workflow to run always but basically noop when no `*.jl` . I'll open a PR with the latter and we can discuss if that is best there.

amontoison:
I just promoted you from `triage` to `write` rights.
Use it wisely ;)

PR #630, opened 2026-07-14
Always run runic and don't directly curl a script from the master branch

apozharski:
Two minor fixes to the `Format.yml` workflow:
1. Always run it. `git-runic` defaults to _only_ checking `*.jl` itself. 
2. Don't do the very scary thing of just curl-ing a script from the tip of a master branch! Instead we now use a tagged release (`v1.7.0`) of `Runic.jl`.

@amontoison I think I like this solution better. the runic workflow is cheap anyway.

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -2,7 +2,6 @@ name: 'Format'
 
 on:
   pull_request_target:
-    paths: ['**/*.jl']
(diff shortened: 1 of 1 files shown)


amontoison (approved):

apozharski:
Ah, this will (smartly) not run the updated workflow because it comes from my fork.

PR #631, opened 2026-07-14
Always run runic and don't directly curl a script from the master branch

apozharski:
Two minor fixes to the `Format.yml` workflow:
1. Always run it. `git-runic` defaults to _only_ checking `*.jl` itself. 
2. Don't do the very scary thing of just curl-ing a script from the tip of a master branch! Instead we now use a tagged release (`v1.7.0`) of `Runic.jl`.

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -2,7 +2,6 @@ name: 'Format'
 
 on:
   pull_request_target:
-    paths: ['**/*.jl']
     types: [opened, synchronize, reopened, ready_for_review]
 
 permissions:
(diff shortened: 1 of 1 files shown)


frapac (approved):

frapac:
Anton, do you know why Runic is stalling? Otherwise, this PR looks good to merge

apozharski:
> Anton, do you know why Runic is stalling? Otherwise, this PR looks good to merge

This is because the runic workflow is triggered on `pull_request_target` and uses the _target_ branch's workflow file which still filters the run.

See the docs here: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

PR #633, opened 2026-07-20
[MadNLPGPU] cuDSS: recover inertia under matching; guard matching off ubatch

michel2323:
cuDSS (through 0.8) stops reporting a usable inertia once matching is enabled: it returns `(0, 0)` with `info == 0` even though the factorization is correct. `InertiaBased`/`InertiaAuto` then bump regularization forever and dive into restoration.

Matching with cuDSS batch solver is not supported.

### DIFF
--- a/lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl
+++ b/lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl
@@ -193,6 +193,25 @@ end
 MadNLP.input_type(::Type{CUDSSSolver}) = :csc
 MadNLP.default_options(::Type{CUDSSSolver}) = CudssSolverOptions()
 MadNLP.is_inertia(M::CUDSSSolver) = (M.inner.matrix.nbatch == 1)  # Uncomment if MadNLP.LU is supported -- (M.opt.cudss_algorithm ∈ (MadNLP.CHOLESKY, MadNLP.LDL))
+
+# Recover the inertia from the sign counts of the factor diagonal D, dumped via the
+# cuDSS "diag" data parameter ("Diagonal of the factorized matrix", i.e. D of P·S·A·S·Pᵀ
+# = L·D·Lᵀ — a congruence of A, so the sign counts equal A's inertia by Sylvester's law,
+# matching scaling/permutation included). Only meaningful for an LDLᵀ factorization,
+# hence the hard check. Exact only for 1×1 pivots, which holds for the (quasi-definite)
+# condensed KKT family this solver targets.
+function inertia_from_diag(M::CUDSSSolver)
+    @assert M.opt.cudss_algorithm == MadNLP.LDL "the factor diagonal D only determines the inertia for an LDLᵀ factorization"
+    n = size(M.tril, 1)
+    CUDSS.cudss_set(M.inner, "diag", M.diag)
+    CUDSS.cudss_get(M.inner, "diag")
+    d = Array(M.diag)
+    z = zero(eltype(d))
+    npos = count(>(z), d)
+    nneg = count(<(z), d)
+    return (npos, n - npos - nneg, nneg)
+end
+
 function MadNLP.inertia(M::CUDSSSolver)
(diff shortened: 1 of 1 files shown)


lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl
```diff
@@ -202,6 +212,21 @@ function MadNLP.inertia(M::CUDSSSolver)
     elseif M.opt.cudss_algorithm == MadNLP.LDL
         # N.B.: cuDSS does not always return the correct inertia.
         if info == 0
+            if M.opt.cudss_matching
```
michel2323, line 234: @frapac @sshin23 @amontoison Is this what we want to do, or do we want to automatically switch to inertia-free if matching is enabled?

frapac: I think your solution is the most appropriate. Are we sure `M.diag` is the diagonal factor?

amontoison: Yes but we should always check that the factorization is an `LDL'` before dumping the diagonal `D`.

lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl
```diff
@@ -47,7 +47,12 @@ function set_cudss_options!(solver::CUDSS.CudssSolver, opt::CudssSolverOptions)
         CUDSS.cudss_set(solver, "pivot_threshold", opt.cudss_pivot_threshold)
     end
     if opt.cudss_matching
-        if pkgversion(CUDSS) < v"0.8"
+        if solver.matrix.nbatch > 1
+            # cuDSS matching (`matching_alg`) fails the *analysis* phase with
+            # CUDSS_STATUS_NOT_SUPPORTED on a uniform-batch solver (through cuDSS 0.8),
+            # so never enable it there (e.g. the two-stage per-scenario batch solver).
+            Base.@warn "cuDSS matching is not supported on uniform-batch (ubatch) solvers; ignoring `cudss_matching = true` for this batched solver." maxlog = 1
```
amontoison, line 54: NVIDIA is using an external library for the matching (HSL MC64).
They probably didn't extended it for batch...

lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl
michel2323: Yes — cuDSS's `"diag"` data parameter returns the "Diagonal of the factorized matrix", i.e. the `D` of `P·S·A·S·Pᵀ = L·D·Lᵀ` (a congruence of `A`, so its sign counts equal `A`'s inertia by Sylvester's law, matching scaling/permutation included).

Verified empirically on an RTX 4080 with cuDSS 0.8 on KKT-like symmetric indefinite matrices (`[H Aᵀ; A -δI]`, n=70 and n=280): the sign counts of the dumped `D` match the eigenvalue-count inertia exactly, both with matching off (where it also agrees with cuDSS's native `"inertia"`) and with matching on (where native reports `(0, 0)`). End-to-end, `SparseCondensedKKTSystem` + `InertiaBased` + `cudss_matching=true` now converges in the identical iteration count/objective as with matching off.

@amontoison addressed in 90102473: the recovery is extracted into `inertia_from_diag` with a hard `@assert cudss_algorithm == MadNLP.LDL`, so the LDLᵀ check travels with the code (previously it was only implied by the enclosing branch).

lib/MadNLPGPU/ext/MadNLPGPUCUDAExt/cudss.jl
michel2323: That matches what we observe: the ubatch analysis phase itself returns `CUDSS_STATUS_NOT_SUPPORTED` as soon as `matching_alg` is set, so the guard just skips matching with a one-time warning instead of failing the whole batched solver.

PR #638, opened 2026-08-11
Update URLs for the madsuite-org rename; codecov upload over OIDC

sshin23:
Follow-up to the organization rename (MadNLP → madsuite-org) and the org-site move to madsuite-org.github.io:

- README badges and links, the Documenter source/deploy targets, and the HybridKKT link in the GPU tutorial point at the new home. GitHub Pages does not redirect across the rename, so the docs links were the ones that had actually broken.
- The codecov badge drops the repo-scoped graphing token and pins the `madsuite-org/MadNLP.jl` slug.
- Coverage uploads switch from the shared `CODECOV_TOKEN` secret to OIDC: the workflow authenticates itself, attribution follows the repo that ran (which also populates codecov's new slug on this PR's first run), and there is no token whose codecov-side binding can go stale after a rename. The same convention is already in place in CNLPModels.jl and cnlpmodels-py.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,7 +12,7 @@ jobs:
   build:
     runs-on:
       - self-hosted
-      - cuda13
+      - cuda
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/install-juliaup@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,6 @@ on:
     - cron: '0 0 * * 0'
 
 permissions:
-  id-token: write
   actions: write
   contents: read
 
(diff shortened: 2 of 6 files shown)


sshin23:
One commit added (`0bbeab18`), reverting only the codecov hunk: the OIDC switch is dropped and the three upload sites stay on `token: ${{ secrets.CODECOV_TOKEN }}` (org-level). Measured while this PR sat open: **OIDC uploads are rejected org-wide** — the mechanism behind the ExaModels/CNLPModels coverage blackout (`Repository not found` behind fail-open green steps) — while the org token works, as this repository's own upload tonight demonstrated (clean, 88.91% posted). The URL, badge, and Documenter changes are untouched and remain the substance of the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #640, opened 2026-08-13
Use MOI.Nonlinear.QPBlockData

blegat:
A net line deletion of 685 lines. Needs

- [ ] https://github.com/jump-dev/MathOptInterface.jl/pull/3048

Two adaptations:

- `MOI's QPBlockData` treats a variable as a parameter if and only if its index is a key of the parameters dictionary, instead of the index-offset convention, so parameters are registered when the constrained variable is added (the value is still re-synced before every solve).
- `eval_constraint_jacobian` and `eval_hessian_lagrangian` now return the number of entries written instead of a one-indexed cursor, so the callers drop the offset adjustment.


frapac (commented):
LGTM! We can merge this PR once the PR has been merged en released in MOI.

frapac (approved):

amontoison:
Please do exactly the sme modifications in `Ipopt.jl` and `MadNLP.jl` such that it is easier to maintain.
I will also do the same modifications in `UnoSolver.jl`.

blegat:
I can take care of Uno as well, I am doing Ipopt first but I plan to do MadNLP the same way.

PR #643, opened 2026-08-19
Use MOI.Nonlinear.ModelWithQuad instead of QPBlockData

blegat:
Proof of concept refactor after

- [x] https://github.com/jump-dev/MathOptInterface.jl/pull/3053

Closes https://github.com/madsuite-org/MadNLP.jl/pull/640


blegat:
Merged into https://github.com/madsuite-org/MadNLP.jl/pull/640

PR #644, opened 2026-08-21
Improvements to actions workflows

apozharski:
No longer use scary permissions additional permissions, and allow PRs from forks to run the runic workflow.


apozharski:
Mostly this is minor maintenance to the workflows introduced in the schur complement PR.

sshin23 (approved):

